PR #2, opened 2015-06-11
Change groupid and artifactid

jomrazek:
We discussed and decided to change groupids and artifactids

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -1,9 +1,9 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jboss</groupId>
-    <artifactId>dependency-analyzer</artifactId>
+    <groupId>org.jboss.da</groupId>
+    <artifactId>parent</artifactId>
     <version>0.0.1-SNAPSHOT</version>
   </parent>
-  <artifactId>dependency-analyzer-rest</artifactId>
+  <artifactId>backend</artifactId>
 </project>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -16,8 +16,8 @@
 
 	<modules>
 
-		<module>dependency-analyzer-backend</module>
-		<module>dependency-analyzer-communication</module>
-		<module>dependency-analyzer-rest</module>
+		<module>backend</module>
```
slaskawi: Maybe a `server` would be a better name?

backend/pom.xml
```diff
@@ -1,9 +1,9 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jboss</groupId>
-    <artifactId>dependency-analyzer</artifactId>
+    <groupId>org.jboss.da</groupId>
```
slaskawi: Shouldn't this be under PNC umbrella? So `org.jboss.pnc.da` ?

jomrazek: We talked about it yesterday and result is this, of course we can still change it easily.

pom.xml
jbartece: I'd say that is much about the preference ;-) . There will be no client, so I wouldn't named it as a server. But it is just my preference :-)

backend/pom.xml
jbartece: It is right, that it is in the PNC space right now, but it is held as a separate project. Hmm. Currently we were speaking about it to held it completely separated - only REST API will be connection point. For now my vote is to keep the groupId separated. 
@matejonnet @rnc Wdyt?

rnc: If we put it under the ``org.jboss.pnc`` groupId then it would be at the same level as the submodules (e.g. SPI) within PNC. This could be regarded as a bit confusing?
I think having it under ``org.jboss.da`` is fine for now (and its not dissimilar to the layout that e.g. https://github.com/droolsjbpm/ uses).

PR #6, opened 2015-06-22
[DA-31] Add logger dependencies

jomrazek:


reports-rest/pom.xml
```diff
@@ -1,9 +1,17 @@
...
+		<artifactId>parent</artifactId>
+		<version>0.0.1-SNAPSHOT</version>
+	</parent>
+	<artifactId>reports-rest</artifactId>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.jboss.logging</groupId>
+			<artifactId>jboss-logging</artifactId>
+		</dependency>
```
rnc: PNC has just standardised on SLF4J (See NCL-908 and https://github.com/project-ncl/pnc/pull/394 ) : perhaps the same should be done here? 

Why the large whitespace change?

jomrazek: Ok, I can look at it and we can change it, whitespace change is caused by formating i don't know how Jakub format it. When i edit what I need I still have changes so i reformat all xmls cause they were inconsistent anyway.

jbartece: I had some weird format setup

jbartece:
Rebase to master please. There are significant changes in the POM files.

PR #7, opened 2015-06-22
Adding initial code for PNC Interface

thescouser89:
[DA-7]

The PNCInterface code will allow DA to talk with PNC.

We are currently using the Resteasy client to do REST calls to PNC.

Properties need to be defined in: `communication/src/main/resources/credentials.properties`
to setup the PNC server location, and keycloak properties.

This is an initial PR to see if I got most things right.

Things I'm unclear/not happy right now:
* How should we define the keycloak and pnc settings?
  - right now it's set in a properties file
  - should we use BASH variables instead?

* Should I allow the user of the `getPNCInstance()` method to specify the location of PNC she wants to use?

In the very near future I'll add the ability to ask PNC to add a new product/project etc. For now I just wanted to get some initial comments/review before moving further.


slaskawi:
Regarding to property file vs BASH properties - I would prefer using a JSON file just like we do in Orchestrator. This way it would be unified.

communication/pom.xml
```diff
@@ -1,9 +1,39 @@
...
+            <version>3.0.11.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-jackson-provider</artifactId>
+            <version>3.0.11.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
```
slaskawi: Is this dep needed? The connectivity should be provided by the RestEasy client.

communication/src/main/java/PNCInterface.java
```diff
@@ -0,0 +1,31 @@
+import org.jboss.da.communcation.pnc.entity.BuildConfiguration;
+import org.jboss.da.communcation.pnc.entity.BuildConfigurationSet;
+import org.jboss.da.communcation.pnc.entity.Product;
+import org.jboss.da.communcation.pnc.entity.Project;
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+public interface PNCInterface {
```
slaskawi: In short-term this is OK, but I would consider small refactoring in Orchestrator and providing an artifact with the contract.

communication/src/main/java/org/jboss/da/communcation/pnc/ReadResource.java
```diff
@@ -0,0 +1,25 @@
...
+    public static String getResource(String key) {
+        InputStream inputStream = ReadResource.class
+                .getClassLoader()
+                .getResourceAsStream("credentials.properties");
+        try {
+            Properties prop = new Properties();
+            prop.load(inputStream);
+            return prop.getProperty(key);
+        } catch (IOException e) {
+            e.printStackTrace();
```
slaskawi: No properties = fatal error IMO.

jbartece:
Maybe it would be good to create a new module for this type of methods. It could be named "common" or something like that

jbartece:
@thescouser89 btw. It is be good to add the JIRA id in the commit msg

communication/src/main/java/PNCInterface.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @GET
+    @Path("pnc-rest/rest/build-configurations")
```
janinko: Shuldn't the "pnc-rest/rest/" part be in target?

communication/src/main/java/org/jboss/da/communcation/pnc/authentication/PNCAuthentication.java
```diff
@@ -0,0 +1,126 @@
...
+        List<BasicNameValuePair> urlParameters = new ArrayList<BasicNameValuePair>();
+        for (String key : urlParams.keySet()) {
+            urlParameters.add(new BasicNameValuePair(key, urlParams.get(key)));
+        }
+        httpPost.setEntity(new UrlEncodedFormEntity(urlParameters));
+        CloseableHttpResponse response = httpclient.execute(httpPost);
+
+        String refreshToken = "";
+        String accessToken = "";
+        try {
```
janinko: Can try-with-resources be used?

communication/src/main/java/org/jboss/da/communcation/pnc/authentication/PNCAuthFilter.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * It works by add the Authorization key, followed by "Bearer <token>"
+     * to the http request
+     *
+     * @param requestContext
+     * @throws IOException
+     */
+    public void filter(ClientRequestContext requestContext) throws IOException {
+        String token = PNCAuthentication.authenticate();
```
janinko: How long the token is valid? Do we need to authenticate again for every request or can we reuse the token?

communication/src/main/java/PNC.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Properties;
+import java.util.ResourceBundle;
+
+public class PNC {
+    /**
+     * Static factory to create an instance of PNCInterface
+     *
+     * @return instance of PNCInterface
+     */
+    public static PNCInterface getPNCInstance(boolean isAuthenticated) {
```
janinko: For usage in container environment, I think this should be producer method (@Produces).

communication/src/main/java/org/jboss/da/communcation/pnc/ReadResource.java
```diff
@@ -0,0 +1,25 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * Created by dcheung on 22/06/15.
+ */
+public class ReadResource {
+
+    public static String getResource(String key) {
```
janinko: IMHO this should be bean and Properties loaded in @PostConstruct method.
I agree with slaskawi that when the properties file doesn't exist, there should be fatal error.

communication/src/main/java/org/jboss/da/communcation/pnc/authentication/PNCAuthFilter.java
pslegr: It depends on the server configuration for tokens validity. As default Keycloak tokens are valid for 5 minutes. It is possible for sure to reuse the access_token, but then you would need to implement the logic to be prepared getting from REST service endpoint response as: "access token expired" and then re-authenticate and try again to access REST service. Of course, refresh tokens are also possibility for getting fresh access_tokens, but it would require enen more additional coding to handle the refreshing tokens.

jbartece:
Rebase to master please. There are significant changes in the POM files. Probably you won't be able to build it, because you don't use the managed dependency versions ;-)

communication/src/main/java/PNCInterface.java
thescouser89: So, the target will become "<pnc-website>/pnc-rest/rest"?

I don't know. I kinda like that I just have to specify the main url in the target and not have to remember to add the `/pnc-rest/rest`. What does @jbartece or @slaskawi think?

communication/src/main/java/org/jboss/da/communcation/pnc/authentication/PNCAuthentication.java
thescouser89: Weeeeeeeeeeee Java-7 stuff. I've been in Java-6 land for far too long :( will refactor!

communication/src/main/java/PNC.java
thescouser89: So that `PNCInterface` can be `@Injected` ? Gotcha!

thescouser89: I'm still thinking in non-containerized Java land :) Thanks for all the input! Will keep that in mind for my next PR!

communication/pom.xml
thescouser89: Heh you are right. This is not needed. Somehow I thought I had to add those to get stuff working cause my IDE wasn't resolving the dependencies.

communication/src/main/java/PNCInterface.java
thescouser89: @slaskawi Do you mean the `Configuration` file in the NCL codebase?
If so, do you think I should also adopt the "@class" and "@module-config" JSON format?

communication/src/main/java/org/jboss/da/communcation/pnc/ReadResource.java
thescouser89: I'll probably remove that file, or rename it to "Configuration". So this class would get PostConstructed with all the properties in it, and then we just have to query it?

thescouser89:
What can I do to get the `@Inject` annotation to work with JUnit tests?

I've refactored the configuration code and moved it to common. Right now it's really simple, and I'm not sure if it should become as fancy as the one in PNC. Any comments welcome :)

communication/src/main/java/org/jboss/da/communcation/pnc/PNCProducer.java
```diff
@@ -0,0 +1,29 @@
...
+import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
+import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
+
+import javax.enterprise.inject.Produces;
+import javax.inject.Inject;
+
+public class PNCProducer {
+    Configuration config = new Configuration();
+    /**
+     * Static factory to create an instance of org.jboss.da.communcation.pnc.PNCInterface
```
janinko: It's not static anymore.

common/src/main/java/org/jboss/da/common/json/DAConfig.java
```diff
@@ -0,0 +1,36 @@
+package org.jboss.da.common.json;
+
+public class DAConfig implements AutoCloseable {
```
janinko: Why add AutoCloseable? There is nothing to close.

communication/src/main/java/org/jboss/da/communcation/pnc/authentication/PNCAuthentication.java
```diff
@@ -0,0 +1,133 @@
...
+        } finally {
+            response.close();
+        }
+        return new String[]{accessToken, refreshToken};
+
+
+    }
+
+    public String authenticate() {
+        try(DAConfig conf = config.getConfig()) {
```
janinko: There is no point for using try-with-resources, as DAConfig shouldn't be closable.

communication/src/main/java/org/jboss/da/communcation/pnc/PNCProducer.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Factory to create an instance of org.jboss.da.communcation.pnc.PNCInterface
+     *
+     * @return instance of org.jboss.da.communcation.pnc.PNCInterface
+     */
+    @Produces
+    public PNC getPNCInstance() {
+        ResteasyClient client = new ResteasyClientBuilder().build();
+
+        try (DAConfig conf = config.getConfig()) {
```
janinko: There is no point for using try-with-resources, as DAConfig shouldn't be closable.

jomrazek:
:+1:

common/src/main/java/org/jboss/da/common/util/Configuration.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private InputStream getConfigStream() throws IOException {
+        String configFileName = System.getProperty(CONFIG_SYSPROP);
+
+        if (configFileName == null) {
+            configFileName = "da-config.json";
+        }
+        // TODO: refactor with proper logger
+        System.out.println("Loading configuration from file: " + configFileName);
```
janinko: So change it to logger or remove it altogether.

janinko:
After rebase and ``System.out.println`` removal :+1:

thescouser89:
See #20 

Closing this one since the branch got a little bit messed up :)

PR #10, opened 2015-06-24
[DA-36] implement REST data model for DA reports

janinko:
As described on **REST endpoints proposal** page.

PR #11, opened 2015-06-24
[DA-22] Create interface for reports generator

janinko:


thescouser89:
Same comment as #10 , should we be using lombok?

PR #20, opened 2015-07-08
Adding PNC interface for Dependency Analysis

thescouser89:
This is just a cleaned up PR with a single commit instead of several commits spread around as in #7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 *~
 */target
 /target/
+*.iml
+.idea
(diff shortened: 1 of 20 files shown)


janinko:
:+1:

jomrazek:
You can solve that dependency error as you write in mail just check that new version doesn't break something.

jbartece:
add to whitelist

jbartece:
retest this please

thescouser89:
Boo jenkins did not catch the error under my test :(

thescouser89:
retest this please

jomrazek:
@thescouser89 You're right for now it is set to skip tests because testsuite is broken so it's testing just if it's possible to build it

PR #22, opened 2015-07-13
Adding initial swagger initialization

thescouser89:
This should enable swagger for reports-rest. Right now we'll have to manually specify which rest class should be scanned for swagger annotations. Will add more code as the PRs for REST endpoints get merged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,6 +140,11 @@
                 <scope>provided</scope>
             </dependency>
             <dependency>
+              <groupId>com.wordnik</groupId>
+              <artifactId>swagger-jaxrs_2.10</artifactId>
+              <version>1.3.12</version>
+            </dependency>
+            <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
                 <scope>test</scope>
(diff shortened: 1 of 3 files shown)


thescouser89:
test me please

jomrazek:
It's throwing Unsatisfied dependencies for type [Set<Service>] with qualifiers [@Default] at injection point [[parameter 1] of [constructor] @Inject com.google.common.util.concurrent.ServiceManager(Set<Service>)] for me while deploying

thescouser89:
Ok it turns out version `1.3.12` was the issue. Version `1.3.11` is fine.

I started to dig further and try to add some swagger annotations, but I have the issue where we define all the annotations on the interface, then have a class that defines the interface. Swagger doesn't like that one bit. Swagger cannot scan the annotations on the interface. So Swagger will miss information like the HTTP verb etc, unless I define the `@Path` annotation both on the interface and on the class implementing the interface.

So right now I'm blocked again. I checked PNC and they don't seem to be using the interface-implementation thing for endpoints. Maybe we could do that also? Not sure if it's worth it or not.

@jomrazek @janinko

thescouser89:
https://github.com/swagger-api/swagger-core/issues/562

thescouser89:
Closing it, will make another commit later.

PR #26, opened 2015-07-15
Update listings.sh so it points to service

janinko:
Changed json parameters so they match new model


janinko:
retest this please

jomrazek:
:+1:

PR #28, opened 2015-07-15
Use ClientRequest instead of resteasy3.

thescouser89:
Previously we were using annotations to setup a REST client (via
resteasy-client) for PNC in Dependency Analysis. This worked well except
when we deploy on JBoss EAP 6.4. JBoss EAP 6.4 uses and provides
resteasy2 libraries, but we need resteasy3 libraries for resteasy-client
to work.

Instead, we have decided to use `ClientRequest` class provided by
resteasy2 to write the REST client. It's a little bit more verbose than
resteasy-client, but it does the job and avoids the version conflict we
have right now.

The `PNC` object will throw some exceptions, if the REST call fails or
the configuration in common cannot be parsed.

The `pom.xml` was adjusted so that we have no more dependency on
resteasy-client, and we use resteasy v2 instead.

I have also decided to rename my 'model classes' package to `model`,
instead of using `entity` to avoid confusion.


thescouser89:
test me please

jomrazek:
:+1:

communication/src/main/java/org/jboss/da/communcation/pnc/PNC.java
```diff
@@ -1,34 +1,68 @@
...
+        Configuration config = new Configuration();
+        DAConfig conf = config.getConfig();
+        pncServer = config.getConfig().getPncServer();
 
-    @GET
-    @Path("products")
-    @Produces(MediaType.APPLICATION_JSON)
-    List<Product> getProducts();
+        PNCAuthentication pncAuthenticate = new PNCAuthentication();
+        token = pncAuthenticate.authenticate();
```
janinko: Maybe the token generation should be in getClient method?

thescouser89: I am scared the `pncAuthenticate.authenticate()` will add lots of latency to the REST calls. So by eagerly loading it, I will eliminate the latency. 

Of course, I haven't really measured how much the latency is :disappointed:  Maybe I should do that.

jbartece: @thescouser89  Can we use the authenticate only if we are not authenticated? (not to spent time on authentication in each request)

jbartece: Great!

janinko: But what about the token expiration?

thescouser89: Woooooooooooooo good point! I've totally missed that. Then maybe getting the token string everytime is the simplest solution for now? What do you think @jbartece ?

Thanks for pointing this out @janinko :smile:

jbartece: Is there an option to check if we are authenticated? But for now it could try to authenticate each time, but create an issue for it please.

thescouser89: Alright right now we'll authenticate everytime.

Related: DA-72

janinko:
:+1:

PR #36, opened 2015-07-22
[DA-28] Implement AProx version finding

jomrazek:


communication/src/main/java/org/jboss/da/communication/aprox/model/VersionResponse.java
```diff
@@ -0,0 +1,59 @@
+package org.jboss.da.communication.aprox.model;
+
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "metadata")
+public class VersionResponse {
+
```
rnc: Would it be worth considering using @jdcasey's Atlas which already has the concept of a ProjectVersionRef ( https://github.com/Commonjava/atlas/tree/master/identities/src/main/java/org/commonjava/maven/atlas/ident/ref) and I believe support for serializing them ?

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.jboss.da.communication.aprox.api.AproxConnector;
+import org.jboss.da.communication.aprox.model.GA;
+import org.jboss.da.communication.aprox.model.GAV;
+import org.jboss.da.communication.aprox.model.GAVDependencyTree;
+import org.jboss.da.communication.aprox.model.VersionResponse;
+
+@Stateless
+public class AproxConnectorImpl implements AproxConnector {
+
```
rnc: Is the AproxConnector interface finalised or will there be other API interface support ?

jbartece:
add to whitelist

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
jomrazek: For current requirements these requests are all we need.

jbartece: It would be better to change it to @ApplicationScoped to use CDI

jbartece:
Please extract code about unmarschalling maven-metadata.xml to single method and create test for it. Also it would be good to create a test for the getVersionsOfGA()

communication/src/main/java/org/jboss/da/communication/aprox/model/VersionResponse.java
jomrazek: John told us that galley supports work with maven metadata but when we consider that we need to download it and parse it I think that this is the simplest way to do it.

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        return null;
+    }
+
+    public GAVDependencyTree getDependencyTreeOfGAV(GAV gav) {
+        // TODO Auto-generated method stub
+        return null;
+    }
+
+    public List<String> getVersionsOfGA(GA ga) {
+        StringBuilder query = new StringBuilder("http://10.3.8.115/api/group/public/"); // update with appropriate aprox group
```
thescouser89: Maybe the ip could be specified in the 'da-config.json' file in the common module?

jomrazek: @thescouser89 yes, this is good idea

PR #40, opened 2015-07-24
[DA-28] Implement AProx version finding

jomrazek:


communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -27,8 +43,52 @@ public GAVDependencyTree getDependencyTreeOfGAV(GAV gav) {
...
+            query.append("/" + ga.getArtifactId() + "/maven-metadata.xml");
+
+            URLConnection connection = new URL(query.toString()).openConnection();
+            tempFile = File.createTempFile("metadata", "xml");
+
+            tempFile = readStream(tempFile, connection);
+
+            return parseMetadataFile(tempFile).getVersioning().getVersions().getVersion();
+        } catch (IOException | ConfigurationParseException e) {
+            e.printStackTrace();
```
jbartece: We should use logging fot that.

common/src/main/resources/da-config.json
```diff
@@ -4,5 +4,6 @@
   "keycloakUsername": "",
   "keycloakPassword": "",
   "keycloakRealm": "",
-  "pncServer": ""
+  "pncServer": "",
+  "aproxServer":"http://10.3.8.115/api/group/public/"
```
jbartece: We sholdn't push private IPs to the public repository

jbartece:
Minor comments, otherwise it looks good. Jozo is on PTO, so we will have to solve it instead of him

PR #41, opened 2015-07-27
Add implementation for DependencyTreeGenerator

thescouser89:
This commit also removes the `GAV` and the `ArtifactTree` classes found
in the backend module, since the exact same model classes are already
implemented in the `communication` aprox models.

~~Note that `SCMLocator` asks for pom path, version, revision, and scm
url. The version string was added since this is needed by the
`AproxConnector` interface.~~

The `AproxConnector.getDependencyTreeOfRevision() ` was changed so that it asks for pom path instead of version.


janinko:
I think the version in AproxConnector.getDependencyTreeOfRevision() is a mistake. There should be path instead.

janinko:
:+1:

PR #42, opened 2015-07-27
Fix bad constructor in PNC class, null values assigned to the fields

jbartece:


janinko:
:+1:

thescouser89:
:+1: Good catch! When I removed the token thing to be set at constructor, I got lazy and forgot to check that the constructor still made sense :( my bad! will add tests to catch this!

PR #45, opened 2015-07-28
Activate Reports JAX-RS class

thescouser89:
Refactored a little bit `ReportsRestActivator.java` so that it's easier to add new JAX-RS classes to the set of classes.

Also added the `Reports.class` class to the set of classes, so that the JAX-RS endpoint `/reports` is activated.


jbartece:
:+1:

janinko:
:+1: Great

PR #51, opened 2015-07-31
Allow `bestMatchVersion` to be null

thescouser89:
Currently, the lombok annotations prevent the `bestMatchVersion` field to be null. This PR changes that behavior so that we can return `null` as JSON value to the key `bestMatchVersion`, instead of getting a `NullPointerException`

JIRA: DA-82, DA-75, DA-76


jbartece:
:+1: Good job!

PR #52, opened 2015-07-31
Remove prints, use logger

jbartece:


reports-backend/src/main/java/org/jboss/da/listings/impl/service/ArtifactServiceImpl.java
```diff
@@ -13,6 +15,8 @@
  */
 public abstract class ArtifactServiceImpl<T extends Artifact> implements ArtifactService<T> {
 
+    private Logger log = LoggerFactory.getLogger(getClass());
```
slaskawi: I would advice using `Methodhandles.lookup().lookupClass()`. This way it's universal (and you might create a template based on that for your IDE).

Does it make sense to have an instance per object? Maybe `private static final` would be better?

jbartece: Yeah, you are right. I'll fix it

slaskawi:
:+1:

PR #56, opened 2015-08-05
[DA-74]  Change the version to OSGi compliant when providing the lookup

jomrazek:


jbartece:
Do not merge this PR for now, please. I have to check some details with Nick.

jbartece:
I found out that the old builds are not OSGi compliant, so if we will change all versions in requests to the OSGi one we won't be able to find the old builds. Right @rnc ? Honza proposed to do it this way: First try to search for it with the original version and if it won't find any match then convert it to the OSGi version and repeat the search.

rnc:
Most EAP and Layered builds should now be OSGi compliant....although you are right in that older artifacts are not.

jbartece:
OK, so we can solve it using the double search as I wrote above.

rnc:
@jdcasey ^ any thoughts on this?

jomrazek:
@jbartece we should change versionGA tests to not depend on external services

jbartece:
@jomrazek Yeah, that's right. Probably we should have two levels of tests - unit tests (no external services) and integration tests (it can use external services)

jbartece:
Rebase to the master and then retest please

jomrazek:
retest this please

jomrazek:
retest this please

jbartece:
We cannot convert the version every time. We have to do it this was: First we use the original version and only if the original version was not found, we will convert version to OSGi one and then we will do the request again

jbartece:
@jomrazek Add a test please

jbartece:
Closing the PR, because jomrazek is on PTO and there is a lot of conflicts, that has to be resolved

PR #64, opened 2015-08-06
[DA-84] implement CLI options for reports

janinko:


thescouser89:
retest this please

cli/bwlist-admin.sh
```diff
@@ -16,6 +16,10 @@ printUsage() {
     echo "    List all artifacts in black or white list"
     echo "$0 pom [--no-transitive]";
     echo "    Check all dependencies from pom in working directory (using dependency:list) and print their Black/White list status"
+    echo "$0 lookup";
+    echo "    Read G:A:Vs from standart input and finds coresponding redhat versions"
```
thescouser89: s/standart/standard
s/coresponding/corresponding

cli/bwlist.sh
```diff
@@ -12,6 +12,10 @@ printUsage() {
     echo "    List all artifacts in black or white list"
     echo "$0 pom [--no-transitive]";
     echo "    Check all dependencies from pom in working directory (using dependency:list) and print their Black/White list status"
+    echo "$0 lookup";
+    echo "    Read G:A:Vs from standart input and finds coresponding redhat versions"
```
thescouser89: s/standart/standard
s/coresponding/corresponding

thescouser89:
I find it a little bit weird that the `lookup` option is the only one that expects values from standard input, whereas for the others we can specify the GAV as a cli parameter.

cli/listings.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-target=10.3.10.85:8080/da/rest/listings
+target=10.3.10.85:8080/da/rest
```
thescouser89: Hardcoded ip address. Not sure what would be a better alternative though.

thescouser89:
I've tried:
```bash
➜  cli git:(d1d1b5e) ✗ echo "xom:xom:1.2.7" | ./bwlist.sh lookup
[DEBUG-dcheung]  [ {"groupId":"xom", "artifactId":"xom", "version":"1.2.7"} ]
Traceback (most recent call last):
  File "./pretty-lookup.py", line 15, in <module>
    if e["blacklisted"] and e["whitelisted"]:
KeyError: 'blacklisted'
➜  cli git:(d1d1b5e) ✗ 
```

Note: I modified `listings.sh` to use my local PNC instance where the lookup endpoint was confirmed to work via swagger-ui.

cli/listings.sh
janinko: I also don't know about better alternative.

jbartece: I think it is not important for now and it will be easier for users to use the tool (they do not need to know the url)

janinko:
> I find it a little bit weird that the lookup option is the only one that expects values from standard input, whereas for the others we can specify the GAV as a cli parameter.

The lookup option is only option that can take multiple GAVs and I'm not fond of having 
```./bwcli.sh lookup G:A:V G:A:V G:A:V G:A:V ...```
Maybe if there would be one optional GAV and if not provided then it would read from stdin?
```./bwlist.sh lookup [GROUP_ID:ARTIFACT_ID:VERSION]
    Finds corresponding redhat versions for Read G:A:V. If the optional parameter is not specified, it reads G:A:Vs from standard input and finds corresponding redhat versions for all of them.

jbartece:
:+1:

thescouser89:
Woops I knew I should have posted before. The lookup command is still throwing some errors, will fix it in another PR!

PR #66, opened 2015-08-07
[DA-61]: Implement the enpoint /reports/gav

jbartece:


jbartece:
retest this please

jomrazek:
:+1:

reports-rest/src/main/java/org/jboss/da/rest/reports/Reports.java
```diff
@@ -48,20 +52,22 @@
...
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "Get dependency report for a GAV")
-    public Report gavGenerator(GAVRequest gavRequest) {
-        // TODO Auto-generated method stub
-        return null;
+    @ApiOperation(
+            value = "Get dependency report for a GAV " // TODO change when dependencies will be implemented
+                    + "(Currently the dependencies and dependency_versions_satisfied don't contains usefull values)")
+    public Report gavGenerator(GAV gavRequest) {
```
jomrazek: I think we should use GAVRequest not GAV here, typo @jbartece ?

jbartece: Currently we do not support products in requests so I removed it. I'm not sure if it is the best approach

PR #68, opened 2015-08-10
[DA-87] bug fix: available versions in report contains null

thescouser89:
This situation occurs if the `bestMatchVersion` is also null. This
occurs when no best match is found.

Previously even if the best match was null, it was also added to the
available versions when calling `setBestMatchVersion` in `ArtifactReport`.

This commit now checks if the best match is *not null* before adding it to
the set of available versions. It still allows `bestMatchVersion` to be
set to null though, since we might have legitimate reasons to set
`bestMatchVersion` to null.

The unit test basically sets the `bestMatchVersion` to null, and we scan
the versions in `getAvailableVersions()` to make sure that no element in
the set is null.


jbartece:
:+1: Thanks Dustin!

PR #73, opened 2015-08-11
Cli fix

thescouser89:
Some tiny fixes (and a rename) to the CLI command line options.


jbartece:
@thescouser89 The availableVersions field should be in the response from the /lookup/gav endpoint and I think it was there. It is bug, that it is missing, so you shouldn't remove it from the CLI scripts. I'll fix the endpoint

jbartece:
@thescouser89  The endpoint is fixed in PR https://github.com/project-ncl/dependency-analysis/pull/74

PR #74, opened 2015-08-11
Fix lookup endpoint with missing filed availableVersions

jbartece:


reports-backend/src/main/java/org/jboss/da/reports/backend/impl/VersionFinderImpl.java
```diff
@@ -61,18 +64,30 @@ public String getBestMatchVersionFor(GAV gav, List<String> availableVersions) {
         return findBiggestMatchingVersion(gav, availableVersions);
     }
 
+    private List<String> getBuiltVerionsFor0(List<String> allVersions) {
```
thescouser89: typo in method name, and it ends with '0'

reports-backend/src/main/java/org/jboss/da/reports/backend/api/VersionFinder.java
```diff
@@ -13,13 +14,23 @@
 public interface VersionFinder {
 
     /**
-     * Finds all RedHat built artifacts (with suffix -redhat) with the same GA
+     * Finds all Red Hat built artifacts (with suffix -redhat) with the same GA
      * 
      * @param gav GroupId and ArtifactId, which specifies the artifact
-     * @return Found built RedHat artifacts with the same GA or null if the GA is not found
-     * @throws CommunicationException when there is a problem with communication with remote services.
+     * @return Found built RedHat artifacts with the same GA or null if the GA was not found
```
thescouser89: s/RedHat/Red Hat

I guess we should do that change to be consistent.

thescouser89:
So I've tried the cli with this PR and everything works fine!

So I guess fix the minor issues I found, then feel free to merge it!

reports-backend/src/main/java/org/jboss/da/reports/backend/impl/VersionFinderImpl.java
jbartece: The 0 is intentional. An extracted method, which contains main part of the logic

reports-backend/src/main/java/org/jboss/da/reports/backend/api/VersionFinder.java
jbartece: Yeah. I'll fix it

reports-backend/src/main/java/org/jboss/da/reports/backend/impl/VersionFinderImpl.java
thescouser89: Gotcha! Hopefully we won't have too many of these methods though! So I guess now just the typo in 'versions'

PR #76, opened 2015-08-11
Add response codes to the report endpoints

jbartece:


reports-rest/.tern-project
```diff
@@ -0,0 +1 @@
+{"libs":["browser","ecma5"],"ide":{}}
```
thescouser89: Is that intentional?

reports-rest/src/main/java/org/jboss/da/rest/reports/Reports.java
```diff
@@ -17,13 +17,18 @@
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
```
thescouser89: Unused import statement

reports-rest/.tern-project
jbartece: No. Sorry for this stupid issue

thescouser89:
:+1:

PR #78, opened 2015-08-11
[DA-93] Add pom-report option

thescouser89:
1. Rename `pom` option to `pom-bw` for `da-cli.sh` and `da-cli-admin.sh`
2. Add the `pom-report` option
3. Make `pretty-lookup.py` accept either a JSON list or a JSON object
   The JSON object is returned for the `/reports/gav` endpoint, and the
   JSON list returned for the `/reports/lookup` endpoint. As such, this
   commit removes the '[' and ']' surrounding the JSON received in the
   `report` function in `listings.sh`, since `pretty-lookup.py` can now
   parse a JSON object too.

4. Make `pretty-lookup.py` search for the `message` key, which indicates
   that something went wrong. If the `message` key is found,
   print the `message` value and exit.

   Note: this assumption is weak and should be improved in the future.
   Note2: this change is required to properly handle the new response code
               in the endpoints.

5. Add parantheses to the `print` function in `pretty-lookup.py` to make
   it work with Python3.

Note: I'm not too happy with the `pom-report` option output, but it works for now. The output is the same as `da-cli.sh report gId:aId:ver`

### DIFF
--- a/cli/da-cli-admin.sh
+++ b/cli/da-cli-admin.sh
@@ -14,8 +14,10 @@ printUsage() {
     echo "    Delete artifact GROUP_ID:ARTIFACT_ID:VERSION from black or white list"
     echo "$0 list (b[lack]|w[hite])"
     echo "    List all artifacts in black or white list"
-    echo "$0 pom [--no-transitive]";
+    echo "$0 pom-bw [--no-transitive]";
     echo "    Check all dependencies from pom in working directory (using dependency:list) and print their Black/White list status"
+    echo "$0 pom-report [--no-transitive]";
+    echo "    Check all dependencies from pom in working directory (using dependency:list) and print their report status"
     echo "$0 lookup";
     echo "    Read G:A:Vs from standard input and finds corresponding redhat versions"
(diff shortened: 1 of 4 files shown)


cli/da-cli-admin.sh
```diff
@@ -23,7 +25,7 @@ printUsage() {
     exit
 }
 
-    
+
```
thescouser89: I just removed some trailing whitespace

cli/da-cli.sh
```diff
@@ -19,7 +21,7 @@ printUsage() {
     exit
 }
 
-    
+
```
thescouser89: Removed trailing whitespace

cli/listings.sh
```diff
@@ -101,11 +100,8 @@ check() {
 
 report() {
     matchGAV $1
-    echo `formatGAVjson`
     tmpfile=`mktemp`
-    echo "[" > $tmpfile # WA until there is pretty print for reports
```
thescouser89: `pretty-lookup.py` can now parse JSON objects on top of JSON list, so this is no longer required.

PR #79, opened 2015-08-12
[DA-89] Improvements to Swagger documentation

thescouser89:
1. All the REST endpoint documentation will now show the HTTP status
   code to expect
2. We can now see the expected JSON object/list we will receive from the
   endpoint
3. A description of the parameters expected in the body of some requests
   is provided.


thescouser89:
I am still unsure on the correct wordings of the documentation, like for the description of the parameters.

But now the model schema for the body in the `reports` endpoint is generated correctly, and we can now also see the JSON object/list to expect as response.

jbartece:
:+1: :+1: :+1:

PR #83, opened 2015-08-12
Some changes in models so that swagger does not get confused

thescouser89:
Swagger gets confused when generating the response class in its UI.
Those fixes will unconfuse swagger.


jbartece:
:+1:

PR #84, opened 2015-08-17
[DA-98][DA-101] CLI changes

thescouser89:
1. Change the `pom-report` and `pom-bw` default behaviour. Now it'll only show top-level dependencies. The user can optionally select to show transitive dependencies by using the `--transitive` flag. (DA-98)
2. User can now optionally specify the directory to where the `pom.xml` to be analyzed is located for `pom-report` and `pom-bw`. (DA-101)

### DIFF
--- a/cli/da-cli-admin.sh
+++ b/cli/da-cli-admin.sh
@@ -19,10 +19,10 @@ printUsage() {
     echo "$0 list (b[lack]|w[hite])"
     echo "    List all artifacts in black or white list"
     echo ""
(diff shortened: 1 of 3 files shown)


thescouser89:
Fitting both commits in a single PR because  both commits are inter-connected.

janinko:
:+1: good! I didn't know that ``pusd`` ``popd`` exists, good to know :)

jbartece:
:+1:

PR #85, opened 2015-08-18
[DA-65] Implement create build configuration from PNC

thescouser89:
Add 'Create Build Configuration' functionality to PNCConnector class.

When POSTing data about a new Build Configuration to PNC, the latter
requires us to not specify the id. As such, the `BuildConfiguration`
class was refactored so that `BuildConfigurationCreate` has all the
fields that in a BuildConfiguration minus the id.

`BuildConfiguration` then extends `BuildConfigurationCreate` and adds
the `id` field.

The three required fields for `BuildConfigurationCreate` for the
creation of Build Configurations to succeed are:

- name
- projectId
- environmentId

These three fields are annotated as `@NonNull`.


janinko:
:+1:

jbartece:
:+1:

PR #87, opened 2015-08-19
[DA-100] Distinguish when GA was not found in AProx and when the comm…

janinko:
…unication failed

When GA is not found DA will return 404 or 206 codes.
When communication fails DA will return 502 code.

### DIFF
--- a/reports-rest/src/main/java/org/jboss/da/rest/reports/Reports.java
+++ b/reports-rest/src/main/java/org/jboss/da/rest/reports/Reports.java
@@ -31,6 +31,10 @@ import com.wordnik.swagger.annotations.ApiOperation;
 import com.wordnik.swagger.annotations.ApiParam;
 import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
+import java.lang.invoke.MethodHandles;
+import org.jboss.da.rest.model.ErrorMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Main end point for the reports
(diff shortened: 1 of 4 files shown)


janinko:
There is problem in AProx: https://github.com/Commonjava/aprox/issues/207

jdcasey:
FWIW, I think traditionally 502 or 503 would be a better response. It lets the caller know it's a communication issue with a remote service on the backend.

janinko:
I updated the code from 500 to 502 as jdcasey suggested. But there is still the issue in AProx, until it is resolved and deployed, the test will fail. So until then let this PR :zzz:

jdcasey:
I've deployed a new instance of AProx on the old NCL VM (10.3.8.115:80) with this fix and the new depgraph api.

jbartece:
retest this please

reports-rest/src/main/java/org/jboss/da/rest/reports/Reports.java
```diff
@@ -102,44 +105,41 @@ public Response gavGenerator(@ApiParam(
...
-            } catch (CommunicationException e) {
-                responseStatus = 206; // Partial Content HTTP code
+                VersionLookupResult lookupResult = versionFinder.lookupBuiltVersions(gav);
+                LookupReport lookupReport = toLookupReport(gav, lookupResult);
+                reportsList.add(lookupReport);
+                if (lookupResult == null) {
+                    responseStatus = Status.PARTIAL_CONTENT;
+                }
+            } catch (CommunicationException ex) {
+                return Response.status(502).entity(ex).build();
```
jbartece: Probably we should return JSON formatted response.

jbartece:
:+1:

PR #89, opened 2015-08-26
[DA-94] Add proper response code for listings endpoint

thescouser89:
Added '404' response code for endpoint '/listings/whitelist/gav' and
'/listings/blacklist/gav' in case the artifact queried is not present.

@jbartece @janinko @jomrazek Do you guys think this is enough or I should add more response codes for the other endpoints as well?

### DIFF
--- a/reports-rest/src/main/java/org/jboss/da/rest/listings/Artifacts.java
+++ b/reports-rest/src/main/java/org/jboss/da/rest/listings/Artifacts.java
@@ -15,6 +15,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


jbartece:
I also think, that we should return 200, when the artifact was inserted successfully (instead of the JSON with success). @thescouser89 @janinko @jomrazek What do you think?

jomrazek:
@jbartece @thescouser89 i will take care of adding black/white artifact because it's related to DA-91

janinko:
@thescouser89 we discussed with @jbartece and @jomrazek and came to conclusion we should keep the json response (along with the proper response codes).

thescouser89:
@janinko @jbartece @jomrazek this PR actually sets the response code + keeps the JSON response. So a `404` will also return `{success: false}`.

Also for addition of an artifact, I haven't touched it, so it should still return `200` with a JSON response.

janinko:
@thescouser89 Ah, sorry, I'm blind probably :D

thescouser89:
@janinko But you've got 4 eyes!!! :eyeglasses:

jomrazek:
:+1:

PR #90, opened 2015-08-28
[DA-91] Same GAV can be both on black and whitelist

jomrazek:


reports-rest/src/main/java/org/jboss/da/rest/listings/Artifacts.java
```diff
@@ -138,6 +152,11 @@ public ContainsResponse isBlackArtifactPresent(@QueryParam("groupid") String gro
     public SuccessResponse addBlackArtifact(
             @ApiParam(value = "JSON object with keys 'groupId', 'artifactId', and 'version'") RestArtifact artifact) {
         SuccessResponse response = new SuccessResponse();
+        if (whiteService.isArtifactPresent(artifact.getGroupId(), artifact.getArtifactId(),
+                artifact.getVersion())) {
+            whiteService.removeArtifact(artifact.getGroupId(), artifact.getArtifactId(),
```
thescouser89: The change looks good! I'm just worried that if between the removal of the artifact from the whitelist, and re-addition to the blacklist, an exception gets thrown and the change to the removal of the artifact from the whitelist is not rolled back (should it be rolled back?)

janinko:
I was thinking if we could make the checks in *ArtifactService? So each BlackArtifactService and WhiteArtifactService would have their own implementation of addArtifactMethod.

Also I think we should return information that an artifact was moved from white to blacklist.

jomrazek:
I will edit it

thescouser89:
:+1: :+1: :+1:

PR #91, opened 2015-08-28
REST fixes

thescouser89:
Fixes: 
- DA-117: pom.xml has dependencies on jackson deps with different versions
- DA-118: /lookup/gav endpoint has duplicate groupid, artifactId, and version fields in the JSON reply

- Use '206' for partial content instead of `Status.PARTIAL_CONTENT` since `Status.PARTIAL_CONTENT` is not defined in EAP 6.4.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -10,8 +10,8 @@
     <packaging>ejb</packaging>
     <dependencies>
         <dependency>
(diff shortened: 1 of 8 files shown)


jomrazek:
:+1:

PR #93, opened 2015-08-28
[DA-119] Rename /lookup/gav endpoint to /lookup/gavs

thescouser89:
Since the '/lookup/gav' endpoint accepts several 'gavs', the endpoint
name should be changed to '/lookup/gavs' to reflect that it can accept
several 'gavs'.

Thanks to Nick Cross (@rnc) for proposing that idea!


jomrazek:
:+1:

PR #95, opened 2015-09-01
[DA-102] Add dependencies to the reports generator

janinko:
Until implemented AproxConnectorImpl.getDependencyTreeOfGAV() will
return empty GAVDependencyTree


thescouser89:
:+1:

PR #103, opened 2015-09-03
[DA-28] Implement getDependencyTreeOfGAV method

thescouser89:
The method is implemented using the `aprox-depgraph-client-java`
dependency. It provides a Java client library to communicate with the
Aprox instance, and get a list of dependencies of any artifact from
Aprox.

This commit uses a SNAPSHOT version of this artifact.


thescouser89:
@jbartece @janinko @jomrazek I wanted to point out that right now we are only showing the dependency relationship in `GAVDependencyTree`, but we could also in the future show parent relationships too (I think that's important). If there is a need for it, I will add it in another PR.

thescouser89:
@psakar I changed the test so that it uses guava instead of hibernate, since guava has less dependencies. Hope you don't mind! :smiley:

jomrazek:
:+1:

PR #108, opened 2015-09-08
Get rid of null and NoGAVInRepositoryException

janinko:
Optional is used at these places:
  Instead of NoGAVInRepositoryException
    AproxConnector returns Optional of GAVDependencyTree instead of
    throwing NoGAVInRepositoryException.

  Instead of null in Best match version
    Where best match version was used passed or stored, Optional is
    used - from VersionFinder, in VersionLookupResult and in
    ArtifactReport.
    In REST model (LookupReport and Report) null is still possible
    beacuse of mapping into JSON.

  Instead of null from ReportsGenerator
    ReportsGenerator returns Optional of ArtifactReport instead of null.

When no versions are found for a GA it means that there are no built
versions for that GA, so we don't have to handle it in special way.

  Where null was returned beacause of missing GA, empty list is now
  returned.

  This have the implication that response code 206 will never be used in
  /lookup/gavs

### DIFF
--- a/communication/src/main/java/org/jboss/da/communication/aprox/NoGAVInRepositoryException.java
+++ b/communication/src/main/java/org/jboss/da/communication/aprox/NoGAVInRepositoryException.java
@@ -1,8 +1,0 @@
-package org.jboss.da.communication.aprox;
-
-public class NoGAVInRepositoryException extends Exception {
-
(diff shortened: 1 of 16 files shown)


janinko:
Replacing #104

jomrazek:
:+1:

PR #110, opened 2015-09-10
[DA-121] Move SCM test to integration tests

jomrazek:


source-code-manager/src/main/java/org/jboss/da/scm/SCMTypeWrap.java
```diff
@@ -0,0 +1,14 @@
+package org.jboss.da.scm;
+
+public class SCMTypeWrap {
+
```
jbartece: This is probably unnecessary. Remove it please

janinko:
:+1:

thescouser89:
:+1:  tested manually.

PR #117, opened 2015-09-15
Rework B&W list so it is easier for user to use it - LIST version

janinko:
In Whitelist we now allow only redhat versions of artifacts.
  If non-redhat version is being added, error 400 is returned.
  If redhat version is being checked, exact match is searched.
  If non-redhat version is being checked, it is converted to OSGi
  version and then all redhad version of this artifact are searched.

In Blacklist we now store non-redhat OSGi versions of artifact.
  If redhat version is being added, the redhat suffix is stripped and
  then converted to OSGi.
  If non-redhat version is being added, it is converted to OSGi version.
  If redhat version is being checked, the redhat suffix is stripped and
  then converted to OSGi.
  If non-redhat version is being checked, it is converted to OSGi version.

This version returns list (empty or with one artifact) when checking if artifact is in blacklist.
```json
{
    "contains":false,
    "found": []
}
```
```json
{
    "contains":false,
    "found": [{
        "groupId":"org.jboss",
        "artifactId":"dependency-analysis",
        "version":"1.2.3.redhat-2"
     }]
}
```

### DIFF
--- a/cli/listings.sh
+++ b/cli/listings.sh
@@ -95,7 +95,8 @@ check() {
     tmpfile=`mktemp`
     get "listings/${color}list/gav?groupid=${groupId}&artifactid=${artifactId}&version=${version}" > $tmpfile
     if grep -q '"contains":true' $tmpfile; then
-        echo "Artifact $groupId:$artifactId:$version is ${color}listed"
+        echo -n "Artifact $groupId:$artifactId:$version is ${color}listed - actual verisions in list: "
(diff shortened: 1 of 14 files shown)


jomrazek:
I didn't see code, but this version of response is better.

jomrazek:
:+1:

jbartece:
:+1: AProx instance is down and it causes the failed tests

PR #121, opened 2015-09-15
[DA-124]: Improve performance of lookup using parallel computing

jbartece:


jbartece:
Jenkins is down. PR tested locally and tests passed

thescouser89:
:+1:

PR #123, opened 2015-09-15
Fix not catching an exception

janinko:


jbartece:
We solved it locally using @ApplicationScoped beans. It is probably better approach

PR #124, opened 2015-09-15
Fix White/Black list services bug caused by Stateless beans

jbartece:
If the exception in Stateless bean is thrown than it is wrapped with the EJB exception.
The CDI beans don't do this.


thescouser89:
retest this please

thescouser89:
:+1:

PR #125, opened 2015-09-15
Fix for PR 124

thescouser89:
\#124 incorporates changes that breaks deletion.

Error is:

    Removing a detached instance org.jboss.da.listings.api.model.BlackArtifact

This commit attempts to fix that error by first merging and re-assigning
the detached entity.

http://stackoverflow.com/a/17505621


jbartece:
@thescouser89 Based on the linked stackoverflow topic, we should use the getReference method. Or am I wrong?

thescouser89:
@jbartece  I tried the `getReference` method but because we are using Generics (T), I can't really pass to it the generic class object instance.

thescouser89:
retest this please

jbartece:
I think we can do it. Check PR: https://github.com/project-ncl/dependency-analysis/pull/126

thescouser89:
#124 has updated fix for it. Closing.

PR #126, opened 2015-09-15
Fix removing instances from database

jbartece:


jbartece:
Commit pushed to PR https://github.com/project-ncl/dependency-analysis/pull/124

PR #128, opened 2015-09-15
[DA-141] OSGi the redhat version before adding it to whitelist

thescouser89:
This is to avoid errors with this case:

1. Add `haha:haha:1.0.redhat-1` to the whitelist
2. Add `haha:haha:1.0` to the blacklist
3. The artifact `haha:haha:1.0.redhat-1` is not moved to the blacklist


jbartece:
@thescouser89 I don't get the point (3) - RH artifacts should not be on blacklist. There should be only community artifacts

janinko:
:+1:

janinko:
Well... after discussion, we decided that we need to discuss it more, so not :+1: yet.

thescouser89:
:panda_face:

janinko:
So the resolution of our discussion is that we will not OSGi versions coming into whitelist, but will need to fix the issues by other means. So sorry, but :-1:

thescouser89:
:cry:

PR #132, opened 2015-09-16
[DA-143] Checking for artifact without parameters cause NPE

jomrazek:
Checking for artifact without parameters cause NullPointerException


jbartece:
:+1:

PR #137, opened 2015-09-22
Ensure mktemp works on OS X

msavy:
Inspired by this SO thread - http://unix.stackexchange.com/a/84980

Slightly surprised you didn't decide to do this in Ruby/Python rather than shell.

### DIFF
--- a/cli/listings.sh
+++ b/cli/listings.sh
@@ -64,7 +64,7 @@ list() {
 deleteGAVFromList() {
     setColor $1
     matchGAV $2
-    tmpfile=`mktemp`
+    tmpfile=`mktemp 2>/dev/null || mktemp -t 'pncl'`
     delete "listings/${color}list/gav" "`formatGAVjson`" > $tmpfile
     if ! grep -q '"success":true' $tmpfile; then
         echo "Error removing $groupId:$artifactId:$version"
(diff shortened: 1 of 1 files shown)


jbartece:
test this please

jbartece:
@msavy Hi, thanks for your PR. The CLI tool is some kind of PoC. In future we will maybe change the implementation language to Python.

janinko:
Won't it be worth to make function that will do ``mktemp 2>/dev/null || mktemp -t 'pncl'`` instead of repeating it everywhere?

msavy:
@janinko I didn't want to disrupt the script too much. Fair point, though. Happy to refactor it you prefer it.

@jbartece Seems to didn't build properly for some reason, but the CI results don't seem to be related. Is this a failure from elsewhere, do you think?

jbartece:
@msavy Yeah, it is issue in Jenkins. This PR can't break the tests. I let the Jenkins test it automatically and the I realized, it is not necessary.

janinko:
@msavy I refactored it into #138, if you are ok with it we can close this and merge the other.

msavy:
@janinko Yep, seems to run equivalently - but I'm not sure if there are other areas where OS X is tripping up, because I see a bunch of python errors (reported in your jira as DA-145).

PR #139, opened 2015-09-22
Fix AproxRemoteTest

thescouser89:
Very recently Aprox started to give us more dependencies for a GAV than
it should, hence creating a failure in our AproxTest. It turns out that
some things have changed in Aprox, and this change is required to make
Aprox give back the correct list of dependencies for a GAV. The fix is
inspired from the `Gist.java` found here:

https://github.com/Commonjava/aprox-stack-examples/blob/master/aprox/aprox-depgraph-examples/src/main/java/org/commonjava/aprox/ex/Gist.java


jbartece:
:+1:

PR #144, opened 2015-09-24
[DA-110] Create REST API for automated BC creation

jomrazek:


bc-rest/src/main/java/org/jboss/da/bc/BuildConfigurationsProduct.java
```diff
@@ -0,0 +1,150 @@
...
+    @Path("/finish-process")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Finish analyse and create BCs", response = FinishResponse.class)
+    @ApiResponses(value = { @ApiResponse(code = 200, message = "Response succesfully generated") })
+    public FinishResponse finishAnalyse(@ApiParam(
+            value = "Complete detail information needed to create BCs") InfoEntity bc) {
+        return getFinishAnalyseResponse();
+    }
+
+    // TODO remove after implementation
```
thescouser89: Should this method still be here? :smiley:

bc-rest/src/main/java/org/jboss/da/bc/model/BuildConfiguration.java
```diff
@@ -0,0 +1,79 @@
+package org.jboss.da.bc.model;
```
thescouser89: You could perhaps re-use the model from the `communication` module.
```
communication/src/main/java/org/jboss/da/communication/pnc/model/BuildConfiguration.java
```

bc-rest/src/main/java/org/jboss/da/bc/BuildConfigurationsProduct.java
jbartece: Yes :-). It is a Mock of data, which will be there until we finish this feature :-)

bc-rest/src/main/java/org/jboss/da/bc/model/BuildConfiguration.java
jbartece: These entities are different. BC in communication is PNC BC, but this BC is our model for BC import process in PNC. The entities are slightly different

jbartece:
:+1:

janinko:
:+1:

PR #146, opened 2015-09-25
[DA-132] Fix for cycle in dependency tree

thescouser89:
The fix is actually not really a fix.

The `GAVDependencyTree` generated is a huge tree where checking if a
cycle happened through its dependencies can be problematic and time
consuming.

Instead, the fix involves keeping a set of `GAVDependencyTree` nodes
already visited once it has been processed in `ReportsGenerator`. This
then allows us to:

- display the GAV and its dependencies the first time it is encountered
- only display the GAV the other subsequent times it is encountered.

This has the effect of:

1. Not getting the `ReportsGenerator` stuck in an infinite loop because
   of cycles in `GAVDependencyTree`

2. Produce skinner reports since we now don't show the dependencies for
   a GAV everytime it in encountered while generating the report, we
   only show it once.


janinko:
:+1:

jbartece:
:+1:

PR #151, opened 2015-10-01
[DA-149]: Check for existing BCs in PNC

jbartece:


jbartece:
retest this please

communication/src/main/java/org/jboss/da/communication/pnc/impl/PNCConnectorImpl.java
```diff
@@ -11,38 +11,48 @@
...
-import javax.enterprise.context.ApplicationScoped;
-
 @ApplicationScoped
 public class PNCConnectorImpl implements PNCConnector {
 
     @Inject
-    private Configuration config;
+    private PNCAuthentication pncAuthenticate;
+
+    private String pncBaseUrl;
```
janinko: final

bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BcCheckerImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public Optional<BuildConfiguration> lookupBcByScm(String scmUrl, String scmRevision)
+            throws Exception {
+        List<BuildConfiguration> foundBcs = pncConnector
+                .getBuildConfigurations(scmUrl, scmRevision);
+        if (foundBcs.isEmpty())
+            return Optional.empty();
+        else
+            return Optional.ofNullable(foundBcs.get(0));
```
janinko: Should it be ofNullable? Could get(0) return null? What if it does return null?

communication/src/main/java/org/jboss/da/communication/pnc/impl/PNCConnectorImpl.java
```diff
@@ -11,38 +11,48 @@
...
+
+    @Deprecated
+    public PNCConnectorImpl() {
+    }
```
janinko: Is this used still?

janinko:
:+1:

bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BcCheckerImpl.java
jbartece: It cannot return null. I'll fix it

communication/src/main/java/org/jboss/da/communication/pnc/impl/PNCConnectorImpl.java
jbartece: Yep

communication/src/main/java/org/jboss/da/communication/pnc/impl/PNCConnectorImpl.java
jbartece: It is there because of the Constructor injection. CDI is complaining when you use constructor injection and you have no constructor without parameters

jomrazek:
:+1:

PR #152, opened 2015-10-05
[DA-142] Initial tests for cli

thescouser89:
Note: for some tests, we only test whether the cli outputs something,
      since we cannot do direct string comparison of expected vs actual
      output as the order that some dependencies appear always change.


jbartece:
retest this please

jbartece:
:+1: 
In future we could deal with changing values in responses using chain of greps of the response. But it looks good as a first version :-)

PR #154, opened 2015-10-06
[DA-144] Implement dynamic loading of remote repositories

jomrazek:


jomrazek:
retest this please

jomrazek:
retest this please

janinko:
retest this please

jomrazek:
retest this please

jbartece:
By the REST conventions the URL should be probably /repositories
and I would propose to refactor the /repositories endpoints to the single class and to move it to the same level as reports.
So, it would be:
/reports/*
/repositories/*

jomrazek:
we was talking that we add these endpoints to reports but np i'll change it

jbartece:
:+1:

PR #155, opened 2015-10-06
[DA-153] Implement getDependencyTreeOfRevision given a GAV

thescouser89:
This is done by adding support for it in the `PomAnalyzer` class.


thescouser89:
@janinko I am unsure whether using `Optional` would make sense here, but I know you'll tell me `Optional` always make sense :smile:

jomrazek:
:+1:

janinko:
``Optional`` always make sens, but :+1:  :wink:

PR #156, opened 2015-10-07
[DA-145][DA-148] Fix for projects analysis with cli

thescouser89:
Some projects' dependencies cannot be analyzed using the 'da-cli.sh'
script (for pom-bw and pom-report) because the `mvn dependency:list`
command fails.

This typically happens when we are missing an artifact which only gets
produced during the package phase.

The fix is to capture the failure and recommend the user to run `mvn
clean install` before running the 'da-cli.sh' script again.

### DIFF
--- a/cli/listings.sh
+++ b/cli/listings.sh
@@ -218,12 +218,19 @@ pom_bw() {
     tmpfile=`gettmpfile`
     pushd "${pom_path}" > /dev/null
     mvn -q dependency:list -DoutputFile=$tmpfile -DappendOutput=true $mvn_opts
-    popd > /dev/null
 
     if [ $? -ne 0 ]; then
         rm $tmpfile
+        echo ""
+        echo ""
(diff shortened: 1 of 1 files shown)


cli/listings.sh
```diff
@@ -218,13 +218,20 @@ pom_bw() {
...
+        echo ""
+        echo ""
+        echo "================================================================="
+        echo "'mvn dependency:list' command failed."
+        echo "Consider running 'mvn clean install' before running the pom-bw command again to fix the issue"
+        echo "================================================================="
         exit
     fi
 
+    popd > /dev/null
```
thescouser89: `popd` has to be moved because it interferes with the exit status check on line 222.

jomrazek:
retest this please

jbartece:
It looks good to me, but it would be good if @janinko could review it ;-)

janinko:
:+1:

PR #157, opened 2015-10-08
[DA-72]: Solve reauthentication to Keycloak, when token expires

jbartece:
New interceptor is created, which solves the situation, when access token expired. When it happens,
it triggers reauthentication and force reexecution of the called method in PNC connector

PNCAuthentication class was reworked to be a bean

Integration test for PNCConnector was created

Beans.xml file in communication was fixed to support interceptors

### DIFF
--- a/communication/src/main/java/org/jboss/da/communication/pnc/api/PNCConnector.java
+++ b/communication/src/main/java/org/jboss/da/communication/pnc/api/PNCConnector.java
@@ -1,12 +1,12 @@
 package org.jboss.da.communication.pnc.api;
 
-import java.util.List;
-
 import org.jboss.da.communication.pnc.model.BuildConfiguration;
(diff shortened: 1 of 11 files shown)


jomrazek:
:+1:

jomrazek:
retest this please

jbartece:
retest this please

thescouser89:
So from what I understand, we do a REST call to PNC with some access token, if the access token is invalid and/or old, PNC will complain about it. If PNC complain about authorization needed, we throw an exception. The exception gets intercepted by the interceptor (heh!) and it renews the access token, before repeating the request again.

@jbartece  Question: Will the interceptor create an infinite loop if say, we keep using invalid access token?

thescouser89:
:+1:

jbartece:
@thescouser89 You understand it correctly. And there is no possible infinite loop. If the connector throws exception second time, than it will be propagated to the caller

PR #160, opened 2015-10-08
BC generator: Search for dependencies in produc repo

janinko:
When dependencies are not found in AProx, try original repo of the
product.


jbartece:
:+1:

PR #158, opened 2015-10-08
Link BC backend to BC rest endpoints start-process and analyse-next-level

janinko:

PR #162, opened 2015-10-08
Throw proper response code for exceptions in bc rest endpoint

thescouser89:


thescouser89:
Related to PR #158

jbartece:
Probably we need rebase here

PR #172, opened 2015-10-14
Edit finish-response and change generated BCSet name

jomrazek:


bc-backend/src/main/java/org/jboss/da/bc/backend/impl/FinalizerImpl.java
```diff
@@ -43,16 +43,17 @@
...
-    @Asynchronous
-    public void createBCs(String name, String productVersion, ProjectHiearchy toplevelBc,
+    public Integer createBCs(String name, String productVersion, ProjectHiearchy toplevelBc,
             String bcSetName) {
         try {
             Set<Integer> ids = create(toplevelBc);
             int productVersionId = bcSetGenerator.createProduct(name, productVersion);
             bcSetGenerator.createBCSet(bcSetName, productVersionId, new ArrayList(ids));
+            return productVersionId;
         } catch (Exception ex) {
```
janinko: I think we should throw the exception now that the method is not asynchronous.

bc-backend/src/main/java/org/jboss/da/bc/backend/impl/FinalizerImpl.java
```diff
@@ -43,16 +43,17 @@
...
             log.error("Failed to finalize produc import process", ex);
         }
+        return null;
```
janinko: I think we should throw the exception now that the method is not asynchronous.

jbartece: OK. I'll fix it

jbartece:
Added fix in https://github.com/project-ncl/dependency-analysis/pull/175

PR #177, opened 2015-10-14
Refactor BuildConfigurationGeneratorImpl into ProjectHiearchyCreator

janinko:
When there is log.warn in ProjectHiearchyCreator we can add some functionality in the future to propagate these errors into frontend
```
entity.addError("it doesn't work!");
```


jbartece:
:100:

PR #181, opened 2015-10-16
Disable SVN test

thescouser89:
Given that we are not supporting svn cloning right now, disable svn
tests for now.

We are also disabling it because from time to time the test fails for no
apparent reason. Reducing the amount of noise in our testsuite will be
beneficial.


thescouser89:
@jbartece Should our PRs go into both the 0.4.x branch and master?

jbartece:
@thescouser89 Yes, I think so. Otherwise we won't be able to test 0.4.2 version before release

jbartece:
retest this please

PR #182, opened 2015-10-20
Check BC name by the PNC regex, fix BC generated name

jbartece:


bc-backend/src/main/java/org/jboss/da/bc/impl/BuildConfigurationGeneratorImpl.java
```diff
@@ -100,6 +103,12 @@ private void validate(ProjectHiearchy hiearchy) throws IllegalStateException {
         if (!project.isUseExistingBc() && project.getProjectId() == null)
             throw new IllegalStateException("Project id is null for " + project.getGav());
 
+        Matcher m = bcNamePattern.matcher(project.getName());
+        if (!m.matches())
+            throw new IllegalStateException(
```
jomrazek: We don't catch this exception in rest startAnalyse endpoint where is this method used.

jbartece: In which endpoint? We catch Exception in the finish-process endpoint

jomrazek:
:+1: github shows it wrong for me

PR #183, opened 2015-10-20
[DA-171] Know if Aprox can analyze a GAV or not.

thescouser89:
Previously, when we asked Aprox to analyze a GAV for its dependencies,
we received the same result whether the GAV was:

- not present in Aprox and therefore cannot be analyzed
- present in Aprox and having no compile dependencies.

This commit fixes it by first checking if Aprox has the requested GAV in
its public repository. If yes, then proceed to the actual analysis.

However if the GAV is not in the public repository, a
`FindGAVDependencyException` is thrown.

Right now the `DependencyTreeGeneratorImpl` just catches the new
exception and returns an `Optional.empty()`. This will be really fixed
in DA-170 in another commit.

The JUnit tests are also adjusted to test/adapt for the new behaviour.

### DIFF
--- a/communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
+++ b/communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
@@ -26,6 +26,9 @@ import org.jboss.da.communication.aprox.model.Repository;
 import org.jboss.da.communication.aprox.model.VersionResponse;
 import org.jboss.da.communication.model.GA;
 import org.jboss.da.communication.model.GAV;
+import org.jboss.da.communication.pom.api.PomAnalyzer;
+import org.jboss.da.communication.pom.model.MavenProject;
+import org.slf4j.Logger;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
@@ -45,12 +48,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
-import org.jboss.da.communication.pom.api.PomAnalyzer;
-import org.jboss.da.communication.pom.model.MavenProject;
-
 @ApplicationScoped
 public class AproxConnectorImpl implements AproxConnector {
 
+    @Inject
+    private Logger log;
+    
     @Inject
     private Configuration config;
 
@@ -58,8 +61,8 @@ public class AproxConnectorImpl implements AproxConnector {
     private PomAnalyzer pomAnalyzer;
 
     @Override
-    public Optional<GAVDependencyTree> getDependencyTreeOfGAV(GAV gav)
-            throws CommunicationException, FindGAVDependencyException {
+    public GAVDependencyTree getDependencyTreeOfGAV(GAV gav) throws CommunicationException,
+            FindGAVDependencyException {
 
         if (!doesGAVExistInPublicRepo(gav)) {
             throw new FindGAVDependencyException("Could not find: " + gav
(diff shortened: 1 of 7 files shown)


reports-backend/src/main/java/org/jboss/da/reports/backend/impl/DependencyTreeGeneratorImpl.java
```diff
@@ -40,7 +41,12 @@ public GAVDependencyTree getDependencyTree(SCMLocator scml) throws ScmException,
 
     @Override
     public Optional<GAVDependencyTree> getDependencyTree(GAV gav) throws CommunicationException {
-        return aproxConnector.getDependencyTreeOfGAV(gav);
+        try {
+            return aproxConnector.getDependencyTreeOfGAV(gav);
+        } catch (FindGAVDependencyException e) {
+            // TODO: better handle this later in DA-170
```
thescouser89: Will be fixed in DA-170

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -82,6 +85,7 @@
             GraphExport export = mod.graph(req);
 
             if (export == null || export.getRelationships() == null) {
+                // no dependencies found
```
janinko: If this mean that the GAV exists but has no dependencies, there shouldn't be ```return Optional.empty()```, but ``` return Optional.of(new GAVDependencyTree(gav))```.

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -61,7 +59,12 @@
 
     @Override
     public Optional<GAVDependencyTree> getDependencyTreeOfGAV(GAV gav)
-            throws CommunicationException {
+            throws CommunicationException, FindGAVDependencyException {
+
+        if (!doesGAVExistInPublicRepo(gav)) {
+            throw new FindGAVDependencyException("Could not find: " + gav
+                    + "in public repo of Aprox");
```
janinko: When the GAV is not found, we should return empty optional.

janinko:
We can distinguish between GAV not present in Aprox (empty optional) and GAV present in Aprox and having no dependencies (``GAVDependencyTree`` with empty ``dependencies``). I think we don't need new exception.

thescouser89:
I don't know. I think that's an implicit assumption we're taking, and that's why I'm not too happy with it.

janinko:
I've created testcase for situation where artifact without dependencies was reported as not found: https://github.com/project-ncl/dependency-analysis/pull/191

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
jbartece: I agree with this approach to throw an exception in this case. It is an exceptional state.

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
jbartece: +1 To return empty GAVDependencyTree. When the export is null we should probably throw a RuntimeException (or is there any case, when it should be null?).

jbartece:
@thescouser89 I added some comments.

janinko:
So then we will have to change to not return ``Optional`` but throw some kind of ``GAVNotFountException``:
``AproxConnector.getPom()``
``ReportsGenerator.getReport()``
``POMInfoGenerator.getPomInfo()``
and maybe at other places too.

thescouser89:
@janinko @jbartece I now return the GAVDependencyTree object (which is empty) if a GAV actually has no dependencies (and added a testcase for it).

@janinko I genuinely don't know what's the best way to do it :cry: . Both `Optional.empty()` and throwing the Exception will work, but I just think that the logic that `Optional.empty()` = GAV not in MRRC is a little bit too implicit and confusing.

In the end, I'll go with what the group decides :smiley: 

@jomrazek Let me know your thoughts too!

thescouser89:
Retest this please

jbartece:
retest this please

janinko:
> I've created testcase for situation where artifact without dependencies was reported as not found: #191

@thescouser89 please don't forgett to unignore the test

jbartece:
@thescouser89 Dustin, merge this PR, please : https://github.com/thescouser89/dependency-analysis/pull/1

I unignored the test and removed Optional from the return type

janinko:
Looks ok

jbartece:
:100:

PR #186, opened 2015-10-22
Fix using existing BCs in BC generator

jbartece:
When the existing BC was found, it was not used, but the application tried to create new one


jomrazek:
:+1:

jbartece:
All tests passed locally

PR #187, opened 2015-10-26
Try to parse the pom.xmls and not fail if we can't

thescouser89:
Some pom.xmls that we analyze have issues which cause the analysis of
_all_ the pom.xmls to fail.

This commit instead attempts to ignore the failed pom.xmls analyzed, and
use only the ones that the analyzer was able to successfully parse.

If a pom.xml was not able to be analyzed, we'll log it.

Unrelated, but this commit also removes some unused imports in
`Reports.java`


jbartece:
It's not a good habit to catch the NPE. You can hide a lot of different issues. You should prevent program to throw the NPE - like in the PR #194

jbartece:
:+1:

PR #188, opened 2015-10-26
[DA-158] Shallow clone git repositories when possible

thescouser89:
Otherwise fallback to the full git cloning.

### DIFF
--- a/source-code-manager/src/main/java/org/jboss/da/scm/impl/SCMClonner.java
+++ b/source-code-manager/src/main/java/org/jboss/da/scm/impl/SCMClonner.java
@@ -11,15 +11,22 @@ import org.apache.maven.scm.provider.git.jgit.JGitScmProvider;
 import org.apache.maven.scm.provider.svn.svnexe.SvnExeScmProvider;
 import org.apache.maven.scm.repository.ScmRepository;
 import org.apache.maven.scm.repository.ScmRepositoryException;
-import javax.enterprise.context.ApplicationScoped;
-import java.io.File;
 import org.apache.maven.scm.command.checkout.CheckOutScmResult;
+import org.slf4j.Logger;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
 
 /**
  * This code is mostly based from the SCM code in pnc-local, written by Ahmed Abu Lawi (@ahmedlawi92)
  */
 @ApplicationScoped
 public class SCMClonner {
+
+    @Inject
+    private Logger logger;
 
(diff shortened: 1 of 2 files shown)


source-code-manager/src/main/java/org/jboss/da/scm/impl/SCMClonner.java
```diff
@@ -56,4 +67,46 @@ private ScmRepository getScmRepository(String scmUrl, ScmManager scmManager)
...
+    private boolean shallowClone(SCMType scmType, String scmUrl, String revision, File cloneTo) {
+
+        // we only support git for shallow cloning
+        if (!scmType.equals(SCMType.GIT)) {
+            return false;
+        }
+
+        try {
+            ProcessBuilder pb;
+            if (revision.isEmpty() || revision == null) {
```
janinko: This could lead to NPE, either switch position of the conditions or use ``StringUtils.isBlank()`` from ``org.apache.commons.lang.StringUtils``

testsuite/src/test/java/org/jboss/da/test/server/scm/SCMTest.java
```diff
@@ -46,6 +46,22 @@ public void shouldCloneGitRepository() throws Exception {
         }
     }
 
+    @Test
+    public void shouldBeAbleToCloneGitTag() throws Exception {
+
+        // this test should use the shallow cloning feature
```
janinko: Do you verify that it was indeed used shallow cloning?

thescouser89: Nah not really.

source-code-manager/src/main/java/org/jboss/da/scm/impl/SCMClonner.java
thescouser89: Ohh good catch!

jbartece:
retest this please

testsuite/src/test/java/org/jboss/da/test/server/scm/SCMTest.java
jbartece: Please use ``git rev-list HEAD --count`` to find out if it is a shallow clone (has to print 1)

thescouser89: :+1:  will do! Great idea

thescouser89: Added the check!

thescouser89:
@jbartece @janinko Added the extra test. Please review! :smile_cat:

jbartece:
:+1:

PR #189, opened 2015-10-26
[DA-170] Status field for bcg endpoint 

thescouser89:
When using the bcg endpoint to analyze dependencies, it would be nice to
also get a status of how the dependencies went. Did the analysis
succeeded, failed, or has not been done yet?

In the longer term, this status field could be used in PNC UI to throw
relevant notifications to the user.

This commit builds on DA-171 which adds a way to know if Aprox could not
analyse a particular GAV. DA-171 is required to know whether the
analysis of the GAV failed, or the GAV had no dependencies. Previously
the two cases mentioned above could not be distinguished.

We add the status field by first creating an enum
(`DependencyAnalysisStatus`), and adding the status field to the backend
DAO `ProjectHierarchy` and the frontend DAO `BuildConfiguration`.

Most of the other changes occur in `ProjectHiearchyCreator` where we use
the new status field to know if we need to analyze a dependency or not.

The status field is populated in `getDependencies` method in
`ProjectHiearchyCreator`.

The rest of the changes is on the methods converting a `ProjectHiearchy`
to a `BuildConfiguration`, and vice-versa.

Thanks to @jsenko for his UI fix on PNC: https://github.com/project-ncl/pnc/pull/685


thescouser89:
Note: this PR includes the DA-171 commit. Please review/merge the #188 PR first before merging this one!

PR #192, opened 2015-10-30
[DA-166][DA-167][DA-169] Check if product-version, bcs and product exists

thescouser89:
bcs = BuildConfigurationSet

Check if the product-version, BuildConfigurationSet and product already exists
in PNC before trying to create a new one.

If there's one that already exist, use that one.

### DIFF
--- a/bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BCSetGeneratorImpl.java
+++ b/bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BCSetGeneratorImpl.java
@@ -10,6 +10,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
 import java.util.List;
+import java.util.Optional;
 
 @ApplicationScoped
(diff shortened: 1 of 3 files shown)


jbartece:
retest this please

jbartece:
I'm not sure about this behavior. E.g. I think we cannot re-use the product version, when the product version already exists. It can differ (and probably will). 
@janinko @jomrazek What do you think about it?

thescouser89:
Hmm I didn't know about that. What would be the current behaviour then?

jbartece:
@thescouser89  Probably we should discuss it on a meeting.

thescouser89:
@jbartece Totally forgot to mention this in the scrum meeting. Will mention it tomorrow!

jomrazek:
@jbartece is there already some output if we should reuse product version?

janinko:
@thescouser89 after the discussion, is this PR still relevant, or should it be closed?

thescouser89:
@janinko I'll close it and create a new one.

PR #196, opened 2015-11-11
[DA-179] bump versions of galley and cartographer

janinko:
Galley in version 0.10.3 fixed issue with NPE.
We had to bump also version of cartographer and add galley-cdi-embedder
so there are no missing or ambiguous dependencies.


jbartece:
:+1:

PR #197, opened 2015-11-12
[DA-171] Know if Aprox can analyze a GAV or not [updated]

jbartece:


jbartece:
PR based on https://github.com/project-ncl/dependency-analysis/pull/183. Created because  @thescouser89 is not available to merge my PR.

jbartece:
retest this please

jomrazek:
:+1:

communication/src/main/java/org/jboss/da/communication/aprox/api/AproxConnector.java
```diff
@@ -23,8 +24,10 @@
      * @param gav
      * @return Optional of dependency tree of GAV
```
janinko: This is false

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -82,9 +84,10 @@
             GraphExport export = mod.graph(req);
 
             if (export == null || export.getRelationships() == null) {
-                return Optional.empty();
+                // no dependencies found
+                return new GAVDependencyTree(gav);
```
janinko: > When the export is null we should probably throw a RuntimeException (or is there any case, when it should be null?).

This doesn't  seem to be solved.

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
```diff
@@ -207,6 +210,44 @@ public RepositoryManipulationStatus removeRepositoryFromGroup(Repository reposit
...
+            query.append(config.getAproxServer());
+            query.append("/api/group/public/");
+            query.append(gav.getGroupId().replace(".", "/")).append("/");
+            query.append(gav.getArtifactId()).append('/');
+            query.append(gav.getVersion()).append('/');
+            query.append(gav.getArtifactId()).append("-").append(gav.getVersion()).append(".pom");
+
+            URLConnection connection = new URL(query.toString()).openConnection();
+            try {
+                connection.getInputStream();
```
janinko: I think this should be in try-with-resources (or close the InputStream in finally)

communication/src/main/java/org/jboss/da/communication/aprox/impl/AproxConnectorImpl.java
jbartece: I was thinking if it is the best approach, because the analysis will fail. What about to return an empty tree and log the warning, that export is null?

jbartece:
Dustin is back, we will continue in the original PR

PR #198, opened 2015-11-12
[DA-170] Status field for bcg endpoint

thescouser89:
When using the bcg endpoint to analyze dependencies, it would be nice to
also get a status of how the dependencies went. Did the analysis
succeeded, failed, or has not been done yet?

In the longer term, this status field could be used in PNC UI to throw
relevant notifications to the user.

This commit builds on DA-171 which adds a way to know if Aprox could not
analyse a particular GAV. DA-171 is required to know whether the
analysis of the GAV failed, or the GAV had no dependencies. Previously
the two cases mentioned above could not be distinguished.

We add the status field by first creating an enum
(`DependencyAnalysisStatus`), and adding the status field to the backend
DAO `ProjectHierarchy` and the frontend DAO `BuildConfiguration`.

Most of the other changes occur in `ProjectHiearchyCreator` where we use
the new status field to know if we need to analyze a dependency or not.

The status field is populated in `getDependencies` method in
`ProjectHiearchyCreator`.

The rest of the changes is on the methods converting a `ProjectHiearchy`
to a `BuildConfiguration`, and vice-versa.

### DIFF
--- a/bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
+++ b/bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
@@ -11,10 +11,12 @@ import org.apache.maven.scm.ScmException;
 import org.jboss.da.bc.backend.api.BcChecker;
 import org.jboss.da.bc.backend.api.POMInfo;
 import org.jboss.da.bc.backend.api.POMInfoGenerator;
+import org.jboss.da.bc.model.DependencyAnalysisStatus;
 import org.jboss.da.bc.model.GeneratorEntity;
 import org.jboss.da.bc.model.ProjectDetail;
 import org.jboss.da.bc.model.ProjectHiearchy;
 import org.jboss.da.communication.CommunicationException;
+import org.jboss.da.communication.aprox.FindGAVDependencyException;
 import org.jboss.da.communication.model.GAV;
 import org.jboss.da.communication.pnc.model.BuildConfiguration;
 import org.jboss.da.communication.pom.PomAnalysisException;
@@ -80,16 +82,12 @@ public class ProjectHiearchyCreator {
         if (!hiearchy.isSelected())
             return; // Not selected, ignoring
 
-        Optional<Set<ProjectHiearchy>> dependencies = hiearchy.getDependencies();
-        if (!dependencies.isPresent())
-            return; // Dependencies were already check with error, ignoring
-
-        if (dependencies.get().isEmpty()) { // Dependencies not yet processed, get and process them
-            GAV gav = hiearchy.getProject().getGav();
-            Optional<GAVToplevelDependencies> deps = getDependencies(gav);
-            hiearchy.setDependencies(toProjectHiearchies(deps));
+        if (hiearchy.getAnalysisStatus().equals(DependencyAnalysisStatus.NOT_ANALYSED)
+                && hiearchy.getDependencies().get().isEmpty()) {
+            // Dependencies not yet processed, get and process them
+            setDependencies(hiearchy);
         } else { // Dependencies already processed, search next level
-            for (ProjectHiearchy dep : dependencies.get()) {
+            for (ProjectHiearchy dep : hiearchy.getDependencies().get()) {
                 iterate(dep);
             }
         }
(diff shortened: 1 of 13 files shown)


thescouser89:
retest this please

bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
```diff
@@ -80,16 +82,12 @@ private void iterate(ProjectHiearchy hiearchy) {
...
-            GAV gav = hiearchy.getProject().getGav();
-            Optional<GAVToplevelDependencies> deps = getDependencies(gav);
-            hiearchy.setDependencies(toProjectHiearchies(deps));
+        if (hiearchy.getAnalysisStatus().equals(DependencyAnalysisStatus.NOT_ANALYSED)
+                && hiearchy.getDependencies().get().isEmpty()) {
+            // Dependencies not yet processed, get and process them
+            setDependencies(hiearchy);
         } else { // Dependencies already processed, search next level
-            for (ProjectHiearchy dep : dependencies.get()) {
+            for (ProjectHiearchy dep : hiearchy.getDependencies().get()) {
```
janinko: I think the analysis status can be FAILED there, so it can throw NoSuchElementException from the ``get()``. There should be probably else if or switch.

reports-backend/src/main/java/org/jboss/da/reports/backend/impl/DependencyTreeGeneratorImpl.java
```diff
@@ -40,13 +40,9 @@ public GAVDependencyTree getDependencyTree(SCMLocator scml) throws ScmException,
...
-    public Optional<GAVDependencyTree> getDependencyTree(GAV gav) throws CommunicationException {
-        try {
-            return Optional.of(aproxConnector.getDependencyTreeOfGAV(gav));
-        } catch (FindGAVDependencyException e) {
-            // TODO: better handle this later in DA-170
-            return Optional.empty();
-        }
+    public Optional<GAVDependencyTree> getDependencyTree(GAV gav) throws CommunicationException,
+            FindGAVDependencyException {
+        return Optional.of(aproxConnector.getDependencyTreeOfGAV(gav));
```
janinko: When it never returns empty optional, I think we can change the return type to just ``GAVDependencyTree``

thescouser89: You are right!

thescouser89: That change is pervasive.... when I change the signature of that method, lots of other stuff changes. See latest commit, and let me know what you think! (I think it's OKish)

bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
```diff
@@ -80,16 +82,12 @@ private void iterate(ProjectHiearchy hiearchy) {
...
-            GAV gav = hiearchy.getProject().getGav();
-            Optional<GAVToplevelDependencies> deps = getDependencies(gav);
-            hiearchy.setDependencies(toProjectHiearchies(deps));
+        if (hiearchy.getAnalysisStatus().equals(DependencyAnalysisStatus.NOT_ANALYSED)
+                && hiearchy.getDependencies().isEmpty()) {
+            // Dependencies not yet processed, get and process them
+            setDependencies(hiearchy);
         } else { // Dependencies already processed, search next level
-            for (ProjectHiearchy dep : dependencies.get()) {
+            for (ProjectHiearchy dep : hiearchy.getDependencies()) {
```
janinko: It will work, but still I feel that it would be better to use
```java
}else if (hiearchy.getAnalysisStatus().equals(DependencyAnalysisStatus.ANALYSED)){
```
Or am I just paranoid?

janinko:
It looks good (other from my paranoid comment) :+1:

bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
thescouser89: :+1: I think you are not paranoid and you make a good point. The status could be 'FAILED' and I won't catch it. Will make the change.

thescouser89:
retest this please

janinko:
:+1:

PR #200, opened 2015-11-17
[DA-186] Add optional repositories field in endpoints

thescouser89:


thescouser89:
@janinko Might also help with DA-176 maybe?

thescouser89:
i_should_write_test

reports-backend/src/main/java/org/jboss/da/reports/backend/impl/DependencyTreeGeneratorImpl.java
```diff
@@ -30,13 +30,13 @@
...
-    SCMConnector SCMConnector;
+    SCMConnector scmConnector;
 
     @Override
     public GAVDependencyTree getDependencyTree(SCMLocator scml) throws ScmException,
             PomAnalysisException {
-        return SCMConnector.getDependencyTreeOfRevision(scml.getScmUrl(), scml.getRevision(),
-                scml.getPomPath());
+        return scmConnector.getDependencyTreeOfRevision(scml.getScmUrl(), scml.getRevision(),
+                scml.getPomPath(), Optional.ofNullable(scml.getRepositories()));
```
janinko: Why not use Empty list when repositories is null? What is the semantic differences between empty optional and empty list?

thescouser89: Nothing really. I could have used an empty list also. Maybe the empty list is better :smiley:

thescouser89: :+1:  will do the change.

reports-backend/src/main/java/org/jboss/da/reports/api/SCMLocator.java
```diff
@@ -24,4 +28,7 @@
     @Getter
     @NonNull
     private String pomPath;
+
+    @Getter
+    private List<String> repositories = Collections.emptyList();
```
janinko: maybe add @NonNull here?

thescouser89: :+1:

thescouser89:
retest this please

jbartece:
retest this please

janinko:
I think it looks OK now.

jbartece:
retest this please

jbartece:
The option to add repositories field should be also in the reports/scm endpoint and I think it should be also optional in every BC in the import process.
Please, create some tests before merging this PR

thescouser89:
@jbartece The only candidate test product I could think of was keycloak (which is why this JIRA started). However, even with the optional repository, we cannot test using keycloak since keycloak has a pom.xml which points to the wrong keycloak version (a SNAPSHOT version).

Officially, that'd be a bug on the keycloak side. But the parent pom.xml disables the problematic pom.xml. Dependency Analyzer has no way yet of detecting disabled pom.xml, it just grabs _any_ pom.xml.

jbartece:
@thescouser89 In this PR it would be good to have it only in the reports/scm endpoint. We will add it to the import process in different PR

thescouser89:
@janinko is a genius. Thank you @jbartece @janinko @jomrazek for helping me with this! :pray: 

I've added the test for keycloak.

jbartece:
Great! :+1:

janinko:
> Btw. This change breaks the backward compatibility. We have to increment the REST API version.

How does adding an optional field break backward compatibility?

jbartece:
In this code it is not optional. But if we make it optional in the REST API, than we don't have to change the version.

thescouser89:
@jbartece  Ah if you don't specify the 'repositories' field, it still works on the REST API. For example, if you look at the client testsuite for dependency-analysis.json, there is no 'repositories' field.

https://github.com/thescouser89/dependency-analysis/blob/da-186-repositories/testsuite/src/test/rest/v-1.0/request/reports/scm/dependency-analysis.json

communication/src/main/java/org/jboss/da/communication/pom/PomAnalyzerImpl.java
```diff
@@ -80,40 +80,47 @@ public GAVDependencyTree readRelationships(File pomRepoDir, File pomPath)
...
+                ProjectVersionRef rootProject = null;
+
                 // for each pom.xml
                 for (Map.Entry<File, ProjectVersionRef> entry : projectVersionRefs.entrySet()) {
+                    // if pomPath is the same as what we found analysis the root repository
+                    if (pomPath.getCanonicalPath().equals(entry.getKey().getCanonicalPath())) {
+                        rootProject = entry.getValue();
+                        break;
+                    }
+                }
```
janinko: can't we just use:
```java
ProjectVersionRef rootProject = projectVersionRefs.get(pomPath);
```

thescouser89: I believe I used the canonical path so that I could compare the absolute paths of both, and make the comparison more robust.

```java
import java.io.File;

public class Main {
    public static void main (String[] args) throws Exception {
        File file = new File("/home/dcheung/test.py");
        File file2 = new File("../../../../../../../../../../../home/dcheung/test.py");

        System.out.println(file.equals(file2));
        System.out.println(file.getCanonicalPath().equals(file2.getCanonicalPath()));
    }
}
```

This will print:
```
false
true
```

jbartece:
Looks good to me :+1: 
@janinko Do you have any objections or we can merge it?

communication/src/main/java/org/jboss/da/communication/pom/PomAnalyzerImpl.java
janinko: :+1:

janinko:
:+1:

jbartece:
@thescouser89 Waiting for Rebase&Merge :-)

thescouser89:
WHO CAUSED CONFLICTS!?

PR #206, opened 2015-11-26
Fix IDE and FindBugs warnings

janinko:


thescouser89:
:+1:  Looks good to me!

PR #209, opened 2015-12-06
[DA-167][DA-169] Behaviour on existing PNC entities

thescouser89:
DA-167: Check if the product version already exists. If it does not,
        create a new one. If it does, re-use the existing one.

DA-169: Check if the BuildConfigurationSet we are trying to use already
        exists. If yes, then throw an error since we assume that the
        BuildConfigurationSet we are going to use hasn't been created
        yet.

The error thrown is `PNCRequestException`.

Note: I didn't write any tests since it's pretty hard to write tests that won't easily break as the BuildConfigurations/BuildConfigurationSet/Product/ProductVersion of the PNC server we'll test with change.

### DIFF
--- a/bc-backend/src/main/java/org/jboss/da/bc/backend/api/BCSetGenerator.java
+++ b/bc-backend/src/main/java/org/jboss/da/bc/backend/api/BCSetGenerator.java
@@ -9,18 +9,36 @@ import java.util.List;
 public interface BCSetGenerator {
 
     /**
-     * Finds if BuildConfigurationSet with specified parameters already exists if not creates it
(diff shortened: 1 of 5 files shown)


thescouser89:
retest this please

janinko:
DA-166 was closed and DA-192 was created. DA-192 is aiming for 0.6.0 and for the time being (for 0.5.0) we want to retain the current behavior - create the product if it doesn't exists. If it exists we can reuse it.

thescouser89:
@janinko will make the change!

jbartece:
:+1:

jbartece:
retest this please

PR #210, opened 2015-12-09
DA-135 Add products into the whitelist

jomrazek:


reports-backend/src/main/java/org/jboss/da/listings/api/dao/ProductDAO.java
```diff
@@ -0,0 +1,100 @@
...
+            throws NoSuchElementException;
+
+    /**
+     * Finds Artifacts of Product with specific name and version
+     * 
+     * @param name
+     * @param version
+     * @return Product with its artifacts
+     */
+    Product findArtifactsOfProduct(String name, String version);
```
janinko: What if the product is not found?

reports-backend/src/main/java/org/jboss/da/listings/api/model/Artifact.java
```diff
@@ -24,12 +26,8 @@
...
     @Getter
     @NonNull
-    private String groupId;
-
-    @Setter
-    @Getter
-    @NonNull
-    private String artifactId;
+    @OneToOne(fetch = FetchType.EAGER)
+    private GA ga;
```
janinko: It isn't One-to-One. One GA can have Many artifacts.

reports-backend/src/main/java/org/jboss/da/listings/impl/dao/ArtifactDAOImpl.java
```diff
@@ -24,19 +27,20 @@ public ArtifactDAOImpl(Class<T> clazz) {
     }
 
     @Override
-    public T findArtifact(String groupId, String artifactId, String version) {
+    public Optional<T> findArtifact(String groupId, String artifactId, String version) {
         try {
             CriteriaBuilder cb = em.getCriteriaBuilder();
             CriteriaQuery<T> cq = cb.createQuery(type);
             Root<T> artifact = cq.from(type);
+            Join<T, GA> ga = artifact.join("ga");
```
janinko: What type of join it is?

reports-backend/src/main/java/org/jboss/da/listings/api/dao/ProductDAO.java
jomrazek: I use method that returns Optional, this method is never used. I'll check why i created this and remove it if there is no point

reports-backend/src/main/java/org/jboss/da/listings/impl/service/ArtifactServiceImpl.java
```diff
@@ -35,7 +35,7 @@ public boolean isArtifactPresent(GAV gav) {
 
     @Override
     public boolean removeArtifact(String groupId, String artifactId, String version) {
-        T artifact = getDAO().findArtifact(groupId, artifactId, version);
+        T artifact = getDAO().findArtifact(groupId, artifactId, version).get();
```
janinko: Get will throw NoSuchElementException when it is empty. In the following if, artifact can't be null.

reports-backend/src/main/java/org/jboss/da/listings/impl/dao/ArtifactDAOImpl.java
jomrazek: inner join

reports-backend/src/main/java/org/jboss/da/listings/api/model/Product.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Setter
+    @Getter
+    @NonNull
+    private String name;
+
+    @Setter
+    @Getter
+    @NonNull
+    private String productVersion;
```
jbartece: I think it doesn't respect the created design: The productVersion and support status  (+ id) should be in the separated entity ProductVersion.

reports-backend/src/main/java/org/jboss/da/listings/api/model/ProductArtifactRelationship.java
```diff
@@ -0,0 +1,26 @@
...
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@NoArgsConstructor(access = AccessLevel.PUBLIC)
+@ToString
+public class ProductArtifactRelationship {
```
jbartece: This class is redundant. It should be removed - it is enough to add a field to the Artifact class with the @ManyToOne annotation

reports-backend/src/main/java/org/jboss/da/listings/api/model/WhiteArtifact.java
```diff
@@ -13,12 +15,23 @@
...
-        super(groupId, artifactId, version);
+    @Setter
+    @Getter
+    private boolean is3rdParty;
+
+    @Setter
+    @Getter
+    @NonNull
+    private String osgiVersion;
+
```
jbartece: Add a field
@ManyToOne
private Product product;

reports-backend/src/main/java/org/jboss/da/listings/api/model/Product.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Setter
+    @Getter
+    private SupportStatus support;
+
+    @Getter
+    @Setter
+    @OneToMany(fetch = FetchType.EAGER)
+    private Set<WhiteArtifact> artifacts;
```
jbartece: To work with the @ManyToOne annotation in the product class add this:
@OneToMany(mappedBy = "product", fetch = FetchType.EAGER)
And probably it would be good to rename it to whiteArtifacts. This name is confusing

reports-backend/src/main/java/org/jboss/da/listings/api/model/Artifact.java
jbartece: Yes, it should be @OneToMany

reports-backend/src/main/java/org/jboss/da/listings/api/model/GA.java
```diff
@@ -0,0 +1,33 @@
...
+    @Setter
+    @Getter
+    @NonNull
+    private String groupId;
+
+    @Setter
+    @Getter
+    @NonNull
+    private String artifactId;
+}
```
jbartece: I'm not 100% sure, but probably there should be:

@OneToMany(mappedBy=..., fetchType=...)
private Set<BlackArtifact> blackArtifacts;

@OneToMany(mappedBy=..., fetchType=...)
private Set<WhiteArtifact> whiteArtifacts;


@janinko @thescouser89 What do you think?

reports-backend/src/main/java/org/jboss/da/listings/api/model/ProductArtifactRelationship.java
jomrazek: You want to have Product field in artifact right? And if you have one artifact in multiple products?

jbartece: You are correct. We need this entity. But it should be @Entity and have an id column. Or do I miss anything?

jomrazek:
retest this please

jbartece:
retest this please

reports-backend/src/main/java/org/jboss/da/listings/api/model/Artifact.java
```diff
@@ -24,12 +25,8 @@
     @Setter
     @Getter
     @NonNull
-    private String groupId;
-
-    @Setter
-    @Getter
-    @NonNull
-    private String artifactId;
+    @ManyToOne(fetch = FetchType.EAGER)
```
janinko: Fetch for @ManyToOne is EAGER by default.

reports-backend/src/main/java/org/jboss/da/listings/impl/service/ArtifactServiceImpl.java
```diff
@@ -35,10 +43,22 @@ public boolean isArtifactPresent(GAV gav) {
...
+            if (artifact.get() instanceof WhiteArtifact) {
+                List<Product> products = productDAO.findProductOfArtifact(groupId, artifactId,
+                        version);
+                for (Product p : products) {
+                    p.removeArtifact((WhiteArtifact) artifact.get());
+                    productDAO.update(p);
+                }
+                getDAO().delete(artifact.get());
+                return true;
+            }
```
janinko: Did you think about implementing this method in BlackArtifactServiceImpl and WhiteArtifactServiceImpl? I think this is typical case of polymorphism.

reports-backend/src/main/java/org/jboss/da/listings/api/model/Product.java
janinko: After some more discussion between jbartece and me, we think it should really be split.
``ProductVersion`` should contain:
```java
    private String productVersion;
    private SupportStatus support;
    @ManyToMany
    private Set<WhiteArtifact> whiteArtifacts;
    @ManyToOne
    private Product product;
```
And ``Product`` will contain just name (for the time being, other fields (BOM URL?) may be added later)

So the relation is one project can have many versions. Each version belongs to single product. (the ``ProductVersion`` represents more release than version string).

reports-backend/src/main/java/org/jboss/da/listings/api/model/WhiteArtifact.java
janinko: There should be:
```java
@ManyToMany
private Product product;
```

janinko:
I see there some naming issues, like findProduct* but returning many products. It should be findProduct**s**

janinko:
I think there are Two approaches possible, but you are mixing them together.
You either:

 * Use @ManyToMany
   * There should not be ``ProductArtifactRelationship``
   * There should be in (the newly created) ``ProductVersion``
   ```java
    @ManyToMany
    private Set<WhiteArtifact> whiteArtifacts;
   ```
   * and in ``WhiteArtifact`` should be:
   ```java
    @ManyToMany
    private Product product;
   ```
   * One of the ``@ManyToMany`` annotations should have ``mappedBy``
 * Use *Relationship entity
   * There should not be ``whiteArtifacts`` in ``ProductVersion`` nor ``product`` in ``WhiteArtifact`` 
   * and in ``ProductArtifactRelationship`` should be:
   ```java
    @ManyToOne
    private WhiteArtifact artifact;
    @ManyToOne
    private Product product;
   ```

reports-backend/src/main/java/org/jboss/da/listings/api/model/GA.java
janinko: I don't think that it should be there.

reports-backend/src/main/java/org/jboss/da/listings/api/model/ProductArtifactRelationship.java
jomrazek: It's not entity hibernate takes care of relationship, this is just a helper class so I can get out of DB pair of product and artifact

jbartece:
retest this please

jomrazek:
retest this please

jomrazek:
retest this please

jbartece:
retest this please

thescouser89:
Apologies for this late comment! I was working on adding CLI support, and I'm a little bit perplexed on the `Delete` endpoint for whitelisted GAV. From what I see, it'll delete all the artifacts that contain this particular GAV?

I'm unsure if we should do that, or if we should specify which artifact to delete (where artifact = GAV and it's associated product).

jomrazek:
@thescouser89 yes it's as you wrote, i was thinking about it and decide to implement it this way, when you want remove artifact from whitelist you want to remove it from all products, but when you make mistake and want to remove it from just one product... yes you have a problem

jbrazdil:
When we want to blacklist an artifact, it must be deleted from all products. But I think we may want to delete an artifact just from one product.

thescouser89:
Another question: not sure if this is an error on my part but trying to update a non-existing whitelisted product seems to fails. I'm expecting to get a 404 error, instead I get a stack trace with error 500.

```
</pre></p><p><b>JBWEB000071: root cause</b> <pre>java.util.NoSuchElementException: No value present
	java.util.Optional.get(Optional.java:135)
	org.jboss.da.listings.impl.dao.ProductVersionDAOImpl.changeProductVersionStatus(ProductVersionDAOImpl.java:39)
```

jbartece:
+1 for adding second method to delete only association WhiteArtifact:ProductVersion.

jomrazek:
@jbartece added, @thescouser89 fixed

thescouser89:
:+1: Fix works! Thanks @jomrazek :smiley:

jbartece:
:+1:

PR #213, opened 2015-12-15
fix wrong reporting of BcError.NO_DEPENDENCY

janinko:


janinko:
@jomrazek I'm not sure why it was at that place or if the new place make sense. Can you look at it and say if it is ok like that or should be moved somewhere else?

bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
```diff
@@ -112,6 +111,7 @@ private void setDependencies(ProjectHiearchy hiearchy) {
             hiearchy.setAnalysisStatus(DependencyAnalysisStatus.ANALYZED);
         } catch (CommunicationException ex) {
             log.warn("Failed to get dependencies for " + gav, ex);
+            hiearchy.getProject().addError(BcError.NO_DEPENDENCY);
             hiearchy.setAnalysisStatus(DependencyAnalysisStatus.FAILED);
         } catch (FindGAVDependencyException ex) {
             ProjectDetail project = toplevel.getProject();
```
jomrazek: You're right, but probably we should add it in the pom and scm exception like we discussed on irc

jomrazek:
:+1:

PR #214, opened 2015-12-21
[DA-210] scm report does not show dependency tree

thescouser89:
Due to commit ae10c3a9, the full depency tree of an SCM URL was not
being shown.

This commit was done so that we could anaylyze the keycloak project,
which contains a pom.xml which cannot be properly parsed. But this
commit had a flawed logic. It should actually analyze all the pom.xmls,
not just the top-level pom. Reason is we have to analyze all the
pom.xmls to get the full dependency tree.

So instead, we'll try our best to parse all the pom.xmls, and if we
can't, we'll just skip it and warn the user about it.


communication/src/main/java/org/jboss/da/communication/pom/PomAnalyzerImpl.java
```diff
@@ -80,54 +80,50 @@ public GAVDependencyTree readRelationships(File pomRepoDir, File pomPath,
...
                 for (Map.Entry<File, ProjectVersionRef> entry : projectVersionRefs.entrySet()) {
-                    // if pomPath is the same as what we found analysis the root repository
-                    if (pomPath.getCanonicalPath().equals(entry.getKey().getCanonicalPath())) {
-                        rootProject = entry.getValue();
-                        break;
+                    Set<ProjectRelationship<?, ?>> relationships;
+                    try {
+                        // find all the dependencies / plugins / parents / plugins etc for each pom.xml
+                        relationships = getDiscoveryResult(tempDir, pomRepoDir, entry.getValue(),
+                                repositories).getAcceptedRelationships();
```
jbartece: @thescouser89 Why do you have to analyse all POMs? It should be enough to analyse the requested module/POM and its dependencies. Or am I miss something?

thescouser89: @jbartece Initially analyzing only the requested module/Pom seemed to make sense. Which is what the previous logic does. But analyzing the `application` folder in dependency-analysis only shows the modules that `application` pom.xml uses, not the modules dependencies... since we never analyzed the modules dependencies.

thescouser89: @jbartece Maybe I wrongly understood what you meant!
Do we actually want to see only the top-level, or the full dependency tree?

jbartece: We want to see the whole tree

thescouser89:
retest this please

jbartece:
retest this please

jbrazdil:
So the ``getDiscoveryResult`` doesn't return dependencies of modules that are dependencies of the root pom? I though that's why there is the maven-like cache of pomfiles (``repoLocations.add(new SimpleLocation("file:" + cacheDir.getAbsolutePath()));`` in ``getRepoLocations``). But I don't know how the cartographer works.

jbartece:
retest this please

jbartece:
retest this please

jbrazdil:
retest this please

thescouser89:
retest this please

thescouser89:
DA-218 will finally test for the response JSON too.

thescouser89:
WHY ARE YOU FAILING NOW!?

jbrazdil:
Maybe rebase?

thescouser89:
retest this please

thescouser89:
Tests were failing simply because our testrun was taking more than 30 minutes. The Jenkins configuration was setup to 'Mark as Aborted' if it took longer than 30 minutes. I've extended it to 60 minutes.

PR #215, opened 2015-12-21
[DA-204] advanced scm report

thescouser89:
REST endpoint is `/reports/scm-advanced`

Behaviour
=========
The new report should be based on the current SCM report and provide
this information:

- whitelisted artifacts
- blacklisted artifacts

- community GAV's, which has bestMatchVersion match community GAV's,

- community GAV's, which has no bestMatchVersion, but have built
  versions of the same GA + if there are whitelisted versions of this
  GA, add them there too

- community GAV's

- The same tree of dependencies as in the SCM report

Note: community === non RH artifacts

Note2: we add both the dependency and the bestMatch version to the
       `communityGavsWithBestMatchVersions`, same for builtVersions +
       whitelisted.

It looks something like this:

```json
{
    "report": {the same report as /reports/scm},
    "blacklistedArtifacts": [],
    "whitelistedArtifacts": [],
    "communityGavsWithBestMatchVersions": [],
    "communityGavsWithBuiltVersions": [],
    "communityGavs": []
}
```

### DIFF
--- a/reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
+++ b/reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
@@ -1,0 +1,117 @@
+package org.jboss.da.reports.api;
+
+import org.jboss.da.common.version.VersionParser;
+import org.jboss.da.communication.model.GAV;
+import org.jboss.da.listings.api.service.WhiteArtifactService;
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+import java.util.HashSet;
+import java.util.Set;
+import lombok.Getter;
+
+public class AdvancedArtifactReport {
+
+    private WhiteArtifactService whiteArtifactService;
+
+    @Getter
+    private ArtifactReport artifactReport;
+
+    @Getter
+    private Set<GAV> blacklistArtifacts = new HashSet<>();
+
+    @Getter
+    private Set<GAV> whitelistArtifacts = new HashSet<>();
+
+    @Getter
+    private Set<GAV> communityGavsWithBestMatchVersions = new HashSet<>();
+
+    @Getter
+    private Set<GAV> communityGavsWithBuiltVersions = new HashSet<>();
+
+    @Getter
+    private Set<GAV> communityGavs = new HashSet<>();
+
+    public AdvancedArtifactReport(ArtifactReport artifactReport,
+            WhiteArtifactService whiteArtifactService) {
+        this.artifactReport = artifactReport;
+        this.whiteArtifactService = whiteArtifactService;
+        advancedReportGeneration();
+    }
+
+    public void advancedReportGeneration() {
+        Set<GAV> modulesAnalyzed = new HashSet<>();
+        populateAdvancedFields(artifactReport, modulesAnalyzed);
+    }
+
+    /**
+     * The logic on how the advanced fields are populated is determined from
+     * DA-204
+     *
+     * @param report          report to analyze
+     * @param modulesAnalyzed set of modules already analyzed
+     */
+    private void populateAdvancedFields(ArtifactReport report, Set<GAV> modulesAnalyzed) {
+
+        for (ArtifactReport dep : report.getDependencies()) {
+            if (modulesAnalyzed.contains(dep.getGav())) {
+                // if module already analyzed, skip
+                continue;
+            } else if (isDependencyAModule(dep)) {
+                // if dependency is a module, but not yet analyzed
+                modulesAnalyzed.add(dep.getGav());
+                populateAdvancedFields(dep, modulesAnalyzed);
+            } else {
+                // only generate populate advanced report with community GAVs
+                if (VersionParser.isRedhatVersion(dep.getVersion())) continue;
+
+                // we have a top-level module dependency
+                if (dep.isWhitelisted()) whitelistArtifacts.add(dep.getGav());
+                if (dep.isBlacklisted()) blacklistArtifacts.add(dep.getGav());
+
+                if (dep.getBestMatchVersion().isPresent()) {
+                    communityGavsWithBestMatchVersions.add(dep.getGav());
+                    // also add bestMatchVersion GAV to set
+                    communityGavsWithBestMatchVersions.add(new GAV(dep.getGroupId(),
+                            dep.getArtifactId(), dep.getBestMatchVersion().get()));
+                } else if (!dep.getAvailableVersions().isEmpty()) {
+                    communityGavsWithBuiltVersions.add(dep.getGav());
+
+                    // also add available versions to set
+                    dep.getAvailableVersions().stream()
+                            .forEach(v -> communityGavsWithBuiltVersions
+                                    .add(new GAV(dep.getGroupId(), dep.getArtifactId(), v)));
+
+                    // add any whitelisted GAVs, if any
+                    addWhitelistedGAVToCommunityGavsWithBuiltVersions(dep.getGroupId(), dep.getArtifactId());
+                } else if (hasWhitelistedGA(dep.getGroupId(), dep.getArtifactId())) {
+                    // no bestmatch version, no built versions, but we have GAVs whitelisted
+                    addWhitelistedGAVToCommunityGavsWithBuiltVersions(dep.getGroupId(), dep.getArtifactId());
+                } else {
+                    communityGavs.add(dep.getGav());
+                }
+            }
+        }
+    }
+
+    private boolean isDependencyAModule(ArtifactReport dependency) {
+        return dependency.getGroupId().equals(artifactReport.getGroupId())
+                && dependency.getVersion().equals(artifactReport.getVersion());
+    }
+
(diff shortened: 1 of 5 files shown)


reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+                if (VersionParser.isRedhatVersion(dep.getVersion())) continue;
+
+                // we have a top-level module dependency
+                if (dep.isWhitelisted()) whitelistArtifacts.add(dep.getGav());
+                if (dep.isBlacklisted()) blacklistArtifacts.add(dep.getGav());
+
+                if (dep.getBestMatchVersion().isPresent()) {
+                    communityGavsWithBestMatchVersions.add(dep.getGav());
+                    // also add bestMatchVersion GAV to set
+                    communityGavsWithBestMatchVersions.add(new GAV(dep.getGroupId(),
```
thescouser89: @jbartece if we don't want this, please remove it!

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+                            dep.getArtifactId(), dep.getBestMatchVersion().get()));
+                } else if (!dep.getAvailableVersions().isEmpty()) {
+                    communityGavsWithBuiltVersions.add(dep.getGav());
+
+                    // also add available versions to set
+                    dep.getAvailableVersions().stream()
```
thescouser89: @jbartece this too!

jbartece:
@thescouser89 Could you add pastebin with the example report generated by this endpoint, please?

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+    private Set<GAV> communityGavs = new HashSet<>();
+
+    public AdvancedArtifactReport(ArtifactReport artifactReport,
+            WhiteArtifactService whiteArtifactService) {
+        this.artifactReport = artifactReport;
+        this.whiteArtifactService = whiteArtifactService;
+        advancedReportGeneration();
+    }
+
+    public void advancedReportGeneration() {
```
jbrazdil: This should be private?

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+    private Set<GAV> communityGavsWithBestMatchVersions = new HashSet<>();
+
+    @Getter
+    private Set<GAV> communityGavsWithBuiltVersions = new HashSet<>();
+
+    @Getter
+    private Set<GAV> communityGavs = new HashSet<>();
+
+    public AdvancedArtifactReport(ArtifactReport artifactReport,
+            WhiteArtifactService whiteArtifactService) {
```
jbrazdil: I don't like providing service in constructor. What is this class? POJO or service (bean)? If POJO, it shouldn't use any services, if it's bean it should inject other beans properly.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+                } else {
+                    communityGavs.add(dep.getGav());
+                }
+            }
+        }
+    }
+
+    private boolean isDependencyAModule(ArtifactReport dependency) {
+        return dependency.getGroupId().equals(artifactReport.getGroupId())
+                && dependency.getVersion().equals(artifactReport.getVersion());
```
jbrazdil: We should decide if an artifact is module by looking into the SCM repository, not by comparsion of groupid/version.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
jbrazdil: If you need it just for the construction of AdvancedArtifactReport, I would pass as a methods parameter, not store it in field.

Or build the AdvancedArtifactReport from ouside, using some service.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+                .anyMatch(white -> white.getArtifactId().equals(artifactId) && white.getGroupId().equals(groupId));
+    }
+
+    private void addWhitelistedGAVToCommunityGavsWithBuiltVersions(String groupId, String artifactId) {
+        if (whiteArtifactService.getAll() == null) {
+            return;
+        }
+        whiteArtifactService.getAll().stream()
+                .filter(white -> white.getArtifactId().equals(artifactId) && white.getGroupId().equals(groupId))
+                .forEach(white -> communityGavsWithBuiltVersions.add(new GAV(groupId, artifactId, white.getVersion())));
```
jbrazdil: I don't thik that every built version is in whitelist. We should use VersionFinder.

jbrazdil: Ah, sorry, I think there is mistake in the issue description or some misunderstanding.
I didn't know there was this part and there is mistake in it:
> * community GAV's, which has no bestMatchVersion, but have built versions of the same GA **+ if there are whitelisted versions of this GA, add them there too**

Shoud be

> * community GAV's, which has no bestMatchVersion, but have built versions of the same GA or if there are whitelisted versions of this GA

So in ``communityGavsWithBuiltVersions`` should only be GAVs that are actual dependencies, no other GAVs from whitelist.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
jbrazdil: We don't want this.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
jbrazdil: We don't want this.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+                } else if (!dep.getAvailableVersions().isEmpty()) {
+                    communityGavsWithBuiltVersions.add(dep.getGav());
+
+                    // also add available versions to set
+                    dep.getAvailableVersions().stream()
+                            .forEach(v -> communityGavsWithBuiltVersions
+                                    .add(new GAV(dep.getGroupId(), dep.getArtifactId(), v)));
+
+                    // add any whitelisted GAVs, if any
+                    addWhitelistedGAVToCommunityGavsWithBuiltVersions(dep.getGroupId(), dep.getArtifactId());
```
jbrazdil: We don't want this.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
```diff
@@ -0,0 +1,117 @@
...
+                } else if (hasWhitelistedGA(dep.getGroupId(), dep.getArtifactId())) {
+                    // no bestmatch version, no built versions, but we have GAVs whitelisted
+                    addWhitelistedGAVToCommunityGavsWithBuiltVersions(dep.getGroupId(), dep.getArtifactId());
```
jbrazdil: As written in https://github.com/project-ncl/dependency-analysis/pull/215/files#r48866954 we just want 
```java
communityGavsWithBuiltVersions.add(dep.getGav());
```

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
thescouser89: I actually asked @jbartece on IRC on Dec 21 and from what I understood, he told me to add both the GAV *and* the alternate versions.

Will remove them I guess then!

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
thescouser89: Makes sense. I'll move the logic to ReportsGeneratorImpl

thescouser89:
Closing it, and opening #216 instead.

reports-backend/src/main/java/org/jboss/da/reports/api/AdvancedArtifactReport.java
thescouser89: Hmm how would you do that? From the folder structure, pom.xml file, or from galley?

PR #216, opened 2016-01-05
[DA-204] advanced scm report (based on comments)

thescouser89:
REST endpoint is `/reports/scm-advanced`

Behaviour
=========
The new report should be based on the current SCM report and provide
this information:

- whitelisted artifacts
- blacklisted artifacts

- community GAV's, which has bestMatchVersion match community GAV's,

- community GAV's, which has no bestMatchVersion, but have built
  versions of the same GA + if there are whitelisted versions of this
  GA, add them there too

- community GAV's

- The same tree of dependencies as in the SCM report

Note: community === non RH artifacts

Note2: we only add the dependency to the
       `communityGavsWithBestMatchVersions`, same for builtVersions +
       whitelisted.

It looks something like this:

```json
{
    "report": {the same report as /reports/scm},
    "blacklistedArtifacts": [],
    "whitelistedArtifacts": [],
    "communityGavsWithBestMatchVersions": [],
    "communityGavsWithBuiltVersions": [],
    "communityGavs": []
}
```

This is a PR based on the comments from: #215

### DIFF
--- a/reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
+++ b/reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
@@ -1,23 +1,29 @@
 package org.jboss.da.reports.impl;
 
 import org.apache.maven.scm.ScmException;
+import org.jboss.da.common.version.VersionParser;
 import org.jboss.da.communication.aprox.FindGAVDependencyException;
 import org.jboss.da.communication.aprox.model.GAVDependencyTree;
 import org.jboss.da.common.CommunicationException;
 import org.jboss.da.communication.model.GAV;
 import org.jboss.da.communication.pom.PomAnalysisException;
+import org.jboss.da.communication.pom.api.PomAnalyzer;
 import org.jboss.da.listings.api.service.BlackArtifactService;
 import org.jboss.da.listings.api.service.WhiteArtifactService;
+import org.jboss.da.reports.api.AdvancedArtifactReport;
 import org.jboss.da.reports.api.ArtifactReport;
 import org.jboss.da.reports.api.Product;
 import org.jboss.da.reports.api.ReportsGenerator;
 import org.jboss.da.reports.api.SCMLocator;
 import org.jboss.da.reports.api.VersionLookupResult;
 import org.jboss.da.reports.backend.api.VersionFinder;
+import org.jboss.da.scm.api.SCM;
+import org.jboss.da.scm.api.SCMType;
 import org.slf4j.Logger;
 
 import javax.inject.Inject;
 
+import java.io.File;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
```diff
@@ -116,4 +126,57 @@ private void addDependencyReports(ArtifactReport ar, Set<GAVDependencyTree> depe
...
+
+    private boolean hasWhitelistedGA(String groupId, String artifactId) {
+        if (whiteArtifactService.getAll() == null) {
+            return false;
+        }
+        return whiteArtifactService.getAll().stream()
+                .anyMatch(white -> white.getArtifactId().equals(artifactId) && white.getGroupId().equals(groupId));
+    }
+
+    private boolean isDependencyAModule(ArtifactReport artifactReport, ArtifactReport dependency) {
```
thescouser89: @janinko Will ask jdcasey if there's a better way to do it from Aprox

jbrazdil: There is ongoing discussion on how to detect what artifacts belong to the same project. Not solved yet.

However, this endpoint works with SCM, so we can look into the SCM repository (eg with ``PomAnalyzer.getPOMFileForGAV``) if the GAV is there and if yes -> it is the same project.

jbartece: :+1: Do it the way with checking if it is in the same SCM and create and issue to find a better way.

jbartece:
retest this please

reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
thescouser89: :+1: great suggestion!

reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
```diff
@@ -68,6 +80,16 @@
     }
 
     @Override
+    public Optional<AdvancedArtifactReport> getAdvancedReportFromSCM(SCMLocator scml) throws ScmException,
+            PomAnalysisException, CommunicationException {
+        Optional<ArtifactReport> artifactReport = getReportFromSCM(scml);
+        // TODO: hardcoded to git
+        // hopefully we'll get the cached cloned folder for this repo
+        File repoFolder = scmManager.cloneRepository(SCMType.GIT, scml.getScmUrl(), scml.getRevision());
```
thescouser89: I'm not very proud of this :disappointed: , but it seems to work with the suggestion.

jbartece:
Looks good to me :+1:

thescouser89:
MERGE MERGE MERGE

PR #217, opened 2016-01-06
Backup of the whitelist and blacklist

jbartece:


common/src/main/java/org/jboss/da/common/util/FileUtils.java
```diff
@@ -0,0 +1,25 @@
+package org.jboss.da.common.util;
```
thescouser89: We already have apache commons io as dependency in our root pom.xml which provides the `deleteDirectory` method.

source-code-manager/src/main/java/org/jboss/da/scm/impl/ScmFacade.java
```diff
@@ -38,9 +44,8 @@ public SCMClonner() {
         scmManager.setScmProvider(SCMType.SVN.toString(), new SvnExeScmProvider());
     }
 
-    public void cloneRepository(SCMType scmType, String scmUrl, String revision, File cloneTo)
+    public void shallowCloneRepository(SCMType scmType, String scmUrl, String revision, File cloneTo)
```
thescouser89: Not sure about this rename :smiley_cat: This method is kind of lying since we do a full clone if the shallow clone fails.

jbrazdil: I would suggest compromise: keep the old name and document it in javadoc. ;-)

reports-backend/src/main/java/org/jboss/da/listings/impl/backup/BackupController.java
```diff
@@ -0,0 +1,123 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Bean, which performs regular backup of the whitelist and blacklist to the remote SCM repository.
+ * When the configuration contains empty URL to the SCM repository then the backup is disabled.
+ * 
+ * @author Jakub Bartecek <jbartece@redhat.com>
+ *
+ */
+@NoArgsConstructor
```
jbrazdil: Why annotate @NoArgsConstructor? Java provides default no-arg constructor.

source-code-manager/src/main/java/org/jboss/da/scm/impl/ScmFacade.java
thescouser89: :+1: Actually I'm even a little bit uncomfortable with the old name. So that sounds like a good compromise

common/src/main/java/org/jboss/da/common/util/FileUtils.java
jbartece: Do we want to depend on this library? I don't like it much because it uses nearly every project and it sometimes happens that there can be compatibility issues. E.g. I solved this issue when 1 library requested Guava 13.0.1 and another Guava 17.x and it was not able to have both of them. So, I use this more defensive approach, when it is easy for me to create method like this.

source-code-manager/src/main/java/org/jboss/da/scm/impl/ScmFacade.java
jbartece: OK. I'll update the javadoc

reports-backend/src/main/java/org/jboss/da/listings/impl/backup/BackupController.java
jbartece: It's a relic. I'll remove it.

common/src/main/java/org/jboss/da/common/util/FileUtils.java
jbrazdil: We are already using appache commons and I don't think replicating code from library is great idea.

jbartece:
retest this please

common/src/main/java/org/jboss/da/common/util/FileUtils.java
jbartece: We had a discussion also with Vaclav in the office and he also said if it is not essential, it is better to avoid these libraries. We use this library only in a few places. So, I'd like to let it as it is and remove the library later.

PR #219, opened 2016-01-07
Disable swagger validator

jbrazdil:
As our service is behing firewall, the external validator can't see our json.

### DIFF
--- a/bc-rest/src/main/webapp/doc/index.html
+++ b/bc-rest/src/main/webapp/doc/index.html
@@ -42,6 +42,7 @@
       }
       window.swaggerUi = new SwaggerUi({
         url: url,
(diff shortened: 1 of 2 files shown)


thescouser89:
I built and deploy this PR, but it didn't seem to change anything for DA-209

jbartece:
Good point Dustin. It's not fix of the DA-209, but it is a fix of an error, which was displayed in swagger. Anyway :+1:  for the fix

thescouser89:
:+1:

jbrazdil:
Yeah, I didn't read DA-209 as I though it was issue we found before Christmas. After I wanted to mark the issue as resolved, I saw my error and changed the title.

PR #220, opened 2016-01-07
[DA-209] Remove @ApiResponse for status 200, add response for other states

jbrazdil:
Respons for status 200 is taken from @ApiOperation, there should not be
@ApiResponse for for status 200.

Also added respons classes for other states than 200.


jbartece:
:+1:

PR #224, opened 2016-01-11
Change isDependencyVersionSatisfied to be true if all dependencies are built regardless the root artifact.

jbrazdil:


jomrazek:
:ok_hand:

PR #233, opened 2016-01-12
Catch exception when there is wrong/weird pom file

jbrazdil:
There was an exception thrown, when pom version contained variable


jomrazek:
:ok_hand:

PR #239, opened 2016-01-14
Da 224

jbrazdil:
First commit is partial, there are TODOs:
>org.jboss.da.reports.impl.ReportsGeneratorImplTest: 3x TODO
>org.jboss.da.rest.reports.Reports: 2x TODO
>org.jboss.da.reports.impl.ReportsGeneratorImpl: 1x TODO
Maybe there will be more to change.

Second commit is actually [DA-224].


jbartece:
Fixed in PR https://github.com/project-ncl/dependency-analysis/pull/245

PR #243, opened 2016-01-14
[DA-222] [0.5.x branch] CLI delete whitelist-product fix

thescouser89:
Delete whitelist-product method in cli requires support status, when it
is not in cli help nor needed by the endpoint.

This commit removes the requirement for status, making it consistent
with the documentation.


jbartece:
We'll cherry-pick this commit to the 0.5 branch.

PR #246, opened 2016-01-15
Fix reports rest version in swagger index

jomrazek:


thescouser89:
:+1: Don't forget to change the hardcoded server url in the other cli scripts!

PR #247, opened 2016-01-18
[DA-228] Remove hardcoded application version, add property loading 

jomrazek:
I can't find out right settings of resource plugin to filter testsuite/src/test/rest/v-1.0/expectedResponse/index.html - i will look at it

I created https://docs.engineering.redhat.com/display/JP/Changing+of+REST+API+version so you know all places where you currently need to change version. We're using constants in app and these constants are used in annotations so I think the best state we can get is changing rest versions on 2 places in constants(for app) and in parent pom(for filter) 

If you know how to read property to annotation let me know.

### DIFF
--- a/bc-rest/src/main/webapp/WEB-INF/web.xml
+++ b/bc-rest/src/main/webapp/WEB-INF/web.xml
@@ -7,11 +7,11 @@
         <servlet-class>io.swagger.jaxrs.config.DefaultJaxrsConfig</servlet-class>
         <init-param>
             <param-name>api.version</param-name>
-            <param-value>0.3.0</param-value>
+            <param-value>${version.bc.rest}.0</param-value>
         </init-param>
         <init-param>
             <param-name>swagger.api.basepath</param-name>
-            <param-value>/da-bcg/rest/v-0.3</param-value>
+            <param-value>/da-bcg/rest/v-${version.bc.rest}</param-value>
         </init-param>
         <load-on-startup>2</load-on-startup>
     </servlet>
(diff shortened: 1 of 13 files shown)


bc-rest/src/main/webapp/WEB-INF/web.xml
```diff
@@ -7,11 +7,11 @@
         <servlet-class>io.swagger.jaxrs.config.DefaultJaxrsConfig</servlet-class>
         <init-param>
             <param-name>api.version</param-name>
-            <param-value>0.3.0</param-value>
+            <param-value>${version.bc.rest}.0</param-value>
```
thescouser89: Do we have to set the version as <>.<>.<>, or can we just use <>.<>  ? I have a feeling the trailing '.0' might bite us  sometimes, but I don't remember anymore what does 'api.version' do.

reports-rest/src/main/webapp/WEB-INF/web.xml
```diff
@@ -7,11 +7,11 @@
         <servlet-class>io.swagger.jaxrs.config.DefaultJaxrsConfig</servlet-class>
         <init-param>
             <param-name>api.version</param-name>
-            <param-value>0.3.0</param-value>
+            <param-value>${version.reports.rest}.0</param-value>
```
thescouser89: Same here :smiley:

jbartece:
I think that Dustin is right, that it should be ideally X.Y.Z, but I checked it and I don't see any easy solution for this except using 4 properties in Maven. So, I'd vote for left it as it is and use only X.Y for the REST API and when we will make any backward incompatible changes, we'll increase the number

jbartece:
No objections, I'm merging

PR #250, opened 2016-01-20
[DA-221] CLI: better handle 5xx error

thescouser89:
The CLI will now try to capture HTTP request errors and print a nice
error message to the user, instead of crashing


jomrazek:
:+1:

jbartece:
:+1:

PR #251, opened 2016-01-25
[DA-198] Do not stop at unselected node in Product import

jbrazdil:


thescouser89:
Formatting issues fixed in #251

PR #252, opened 2016-01-27
[DA-202] Investigate why lookup test is non deterministic

thescouser89:
This non-deterministic behaviour, where the test:

Test RestApiReportsRemoteTest::testGavLookupList

fails might be due to the fact that we are using a parallel stream to
concurrently add values to an ArrayList. However, an ArrayList is not
thread-safe by default and will probably suffer from a race condition,
since addition to the ArrayList is not synchronized.

So instead the ArrayList is replaced by a CopyOnWriteArrayList, which is
thread-safe.


thescouser89:
Ran a quick grep on the repository to see if we use 'parallelStream' anywhere else, and the only other place is in `PNCRemoteTest`, and it looks thread-safe :+1:

jomrazek:
:ok_hand:

jbartece:
I should confess. I created thi bug :-(. +:100:  Dustin

PR #255, opened 2016-02-02
[DA-192] Move creating of products to UI in Product Import

thescouser89:
Instead of trying to create a product, we'll just assume that the
product already exists.

The REST call to the import endpoint doesn't change though. We'll still
have to send the product name.

We'll still try to create a product version if it hasn't been created
yet.

### DIFF
--- a/bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BCSetGeneratorImpl.java
+++ b/bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BCSetGeneratorImpl.java
@@ -41,27 +41,16 @@ public class BCSetGeneratorImpl implements BCSetGenerator {
     }
 
     @Override
-    public Integer createProduct(String name, String productVersion) throws CommunicationException,
-            PNCRequestException {
-        Optional<Product> p = pnc.findProduct(name);
+    public Integer createProductVersion(int productId, String productVersion)
+            throws CommunicationException, PNCRequestException {
 
-        Product product;
-
-        // create product if it does not exist
-        if (!p.isPresent()) {
-            product = pnc.createProduct(new Product(name));
-        } else {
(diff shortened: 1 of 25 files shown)


thescouser89:
@jbartece I've decided to keep the REST endpoint so that you still have to send the product name as a string, instead of sending the product id, so that it is somehow still compatible with the old product import process. Let me know if you'd like me to change it to productId though! I honestly have no preference between those two (productId vs productName)

jbartece:
@thescouser89 I'd change it as we dealt not to maintain what we don't need. So, expect only productId and do not create a product.

bc-backend/src/main/java/org/jboss/da/bc/backend/impl/BCSetGeneratorImpl.java
```diff
@@ -41,17 +41,17 @@ public BuildConfigurationSet createBCSet(String name, Integer productVersionId,
...
 
         Product product;
 
         // create product if it does not exist
         if (!p.isPresent()) {
-            product = pnc.createProduct(new Product(name));
+            product = pnc.createProduct(new Product(productName));
         } else {
-            product = p.get();
+            throw new PNCRequestException("The product selected doesn't exist!");
```
thescouser89: @thescouser89 what the heck was that? You got the logic totally inverted.

jbrazdil:
:+1:

jbartece:
:+1:

jbartece:
retest this please

thescouser89:
retest this please

thescouser89:
Alright now I'm hitting the right testsuite failures :) Will fix the testsuite soon!

thescouser89:
:heart:

jbartece:
:+1:

PR #256, opened 2016-02-05
Fix uncorrected checking for missing entity.

jbartece:
If an error occurs, an exception is thrown, that entity was already read


jbrazdil:
:+1:

PR #261, opened 2016-02-15
Clean and improve code

jbrazdil:

PR #265, opened 2016-02-19
Refactor PomAnalyzerImpl and fix [DA-239] 

janinko:
Some functionality from PomAnalyzerImpl was trasnfered into GalleyWrapper.
Construction of DependencyTree was recreated and split into DependencyTreeBuilder.

Dependency on SCM was removed (SCM operation should be handled by caller methods).

### DIFF
--- a/communication/src/main/java/org/jboss/da/communication/aprox/model/GAVDependencyTree.java
+++ b/communication/src/main/java/org/jboss/da/communication/aprox/model/GAVDependencyTree.java
@@ -2,8 +2,16 @@ package org.jboss.da.communication.aprox.model;
 
 import org.jboss.da.communication.model.GAV;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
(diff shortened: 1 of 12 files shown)


janinko:
All but last commits are from #261, I will rebase this commit when #261 is merged.

janinko:
@thescouser89 @jomrazek Can you review this?

thescouser89:
retest this please

thescouser89:
@janinko  @jbrazdil Yeah you'll probably have to refactor the JSON tests cause the output probably changed.

thescouser89:
@janinko  I played with it and it works! How did you solve the 'don't expand the dependency tree' several times for the same GAV?

What I'm scared is that the place where you'll expand the dependency tree will be different each time, which makes the JSON test impossible to pass.

janinko:
@thescouser89 Yeah, there is some room to improvement, but I was getting desperate and didn't want to spend my youth on it.

The increased analysis time is unfortunate, but I'm not sure if we can improve it much. What could be done:
 * Now I get dependencies using galley even when the dependencies are excluded. This could be resolved by some additional work.
 * Maybe instead of resolving dependencies recursively on our side, we could identify "border" (top-level deps of given project) dependencies and ask ~~AProx~~ Indy for dependency trees of these border dependencies.

>What I'm scared is that the place where you'll expand the dependency tree will be different each time, which makes the JSON test impossible to pass.

Yes it is now non-deterministic and suboptimal (the tree now can be deeper than it should be). I didn't want to make too big changes outside PomAnalyzerImpl, so I made some compromise. Mabye we could add some sorting in the GAVDependencyTree pruneLeaves method that will make it deterministic.

janinko:
I've added few lines (sorting of dependencies) in atempt to make GAVDependencyTree prune deterministic.

thescouser89:
@janinko we could disable those JSON checks for now, and just check if the requests work correctly,
and create a JIRA to fix it in the future. WDYT?

janinko:
@thescouser89 I fixed the JSON test that broke before, but broke another test in the process. I'll try to fix this one now (and hopefuly not break another one).

janinko:
Shuld be fixed, Jenkins do not dissapoint me!

PR #274, opened 2016-04-14
[DA-262] Update aprox client

jomrazek:

PR #275, opened 2016-04-15
Refactor BC object for PNC >= 0.9

thescouser89:
The BuildConfiguration JSON object provided by PNC for creating new BuildConfiguration has changed.

The productVersionId and scmRepoUrl fields have become mandatory, and we can only associate a buildConfiguration with only one productVersion. This change is required for #274 to pass its testsuite.


thescouser89:
Note: This commit should not be merged to the v0.8 branch. It only applies for PNC >= 0.9

PR #280, opened 2016-04-28
[NCL-1930] [DA] Failed to resolve artifact from build-untested group

jomrazek:


communication/src/main/java/org/jboss/da/communication/pom/GalleyWrapper.java
```diff
@@ -49,6 +53,8 @@
 
     private final LocalRepo localRepo;
 
+    private Configuration config = new Configuration();
```
janinko: I don't think it's a good idea to create a been by new. Now it works, but if we change Configuration so it Injects anything, it would break.

I'm not sure what is the best way to approach this, I have two ideas:
 * create field ``String indyReporitory`` and create the URL in constructor (where you add a ``Configuration`` paramenter)
 * add ``Configuration`` paramenter to ``addLocations`` method.

jomrazek:
@janinko ok, I edited it

communication/src/main/java/org/jboss/da/communication/pom/GalleyWrapper.java
```diff
@@ -131,7 +135,7 @@ private void fillModules(Set<Artifact> modules, Artifact artifact) {
      * @param artifact
      * @return MavenPomView of the artifact
      * @throws org.jboss.da.communication.pom.PomAnalysisException
-     * @see #addCentralLocation()
+     * @see #addLocations()
```
janinko: #addDefaultLocations()

PR #293, opened 2016-05-20
[NCL-1891] Add endpoint for analyzing built artifacts

jomrazek:


reports-rest/src/main/java/org/jboss/da/rest/reports/Reports.java
```diff
@@ -219,6 +223,43 @@ public Response alignReport(AlignReportRequest request) {
...
+        if (pomPath == null || pomPath.isEmpty()) {
+            pomPath = "pom.xml";
+        }
+        SCMLocator locator = new SCMLocator(request.getScmUrl(), request.getRevision(), pomPath,
+                request.getAdditionalRepos());
+        try {
+            Set<BuiltReportModule> builtReport = reportsGenerator.getBuiltReport(locator);
+            return Response.status(Status.OK).entity(toBuiltReport(builtReport)).build();
+        } catch (ScmException | PomAnalysisException | CommunicationException e) {
+            log.error("Exception thrown in scm endpoint", e);
```
jbartece: This log should be probably changed

jbartece:
Looks good to me. :+1: 
As @janinko  is not here, merging

jbartece:
Sorry, I hit wrong button

PR #301, opened 2016-06-01
Add logging to Indy connector

janinko:


jomrazek:
:+1:  if we have enough log space

PR #334, opened 2016-08-11
[NCL-1983] Allow SCMImpl to clone different repositories in parallel

stanlyDoge:


source-code-manager/src/main/java/org/jboss/da/scm/impl/SCMImpl.java
```diff
@@ -30,59 +27,46 @@
...
-    @Schedule(hour = "*", minute = "*/10")
-    public void invalidateCache() {
-        Iterator<Map.Entry<SCMSpecifier, DirectoryReference>> it = cache.entrySet().iterator();
-        while (it.hasNext()) {
-            DirectoryReference ref = it.next().getValue();
-            if (!ref.check()) {
-                it.remove();
+        } else {
+            try {
+                return fref.get(30, TimeUnit.MINUTES).get().orElseThrow(()->new IllegalStateException("now completed reference has empy file"));
```
janinko: Typo.

PR #350, opened 2016-09-20
Add workaround for getting id of new build configurations

janinko:
Because creating BuildConfigurations using BPM process doesn't return id
of the newly createted BuildConfiguration, we need to get the id in some
other way. For now we will poll PNC until the BC is created, in the
future we should use websockets to retrieve a notification about BC
being created.


jsenko:
Code seems to be OK as a quick fix, but refactor the thread blocking ASAP or you can run out of application threads.

PR #356, opened 2016-09-26
Cli fix

jbartece:


jbartece:
@janinko  Merging without waiting for core-review as you're out, it's on master and we need the fix asap.

PR #371, opened 2016-10-25
Implement more efficient getToplevelDependencies(SCMLocator)

janinko:


jbartece:
@janinko If I understand it correctly, it should fix this issue, right? NCL-2454

janinko:
@jbartece nope, this is not for GAV but for SCM

PR #380, opened 2016-11-14
[NCL-2519] Add field validation for Buildscript in Product Import Process

stanlyDoge:


janinko:
Sorry, I've reimplement Validation a bit in #389. Could you check it?

PR #385, opened 2016-11-23
Change REST version to 1

janinko:


jsenko:
Don't forget to notify me after merge please, in case i miss it.

PR #394, opened 2016-12-01
Relax GAV validation, allow dash in group id

janinko:
Although maven specification prohibits it, maven doesn't enfore it.
That means considerable amount of project with dash in groupId exists.


janinko:
retest this please

PR #395, opened 2016-12-01
Updated DA doc

stanlyDoge:


docs/users-documentation.html
```diff
@@ -145,16 +133,16 @@
             <li>
             <p>
                 The easiest way to make the CLI tools running is to
-                <a href="https://github.com/project-ncl/dependency-analysis/raw/master/cli-wrap/da-cli-tools-0.8.1-pack.zip">download the ZIP archive</a>
-                from GitHub, which contains all the required support scripts and both CLI tools. The
-                latest versions of the tools are always placed in the folder /cli. When you download the
+                <a href="https://github.com/project-ncl/dependency-analysis/raw/master/cli-wrap/da-cli-tools-1.0.0-pack.zip">download the ZIP archive</a>
```
janinko: The url shoul probably point to (not yet existing) tag 1.0.0. Or the 1.0.x branch

docs/users-documentation.html
```diff
@@ -164,16 +152,37 @@
...
                         </ol>
                 </p>
             </li>
+            
+            <li>
+                After unpacking/clonning CLI tool, you need to configure servers.
+            <p>
+                You have to specify DA server and KeyCloak authentication server. DA server can be set as environmental variable ($ export DA_SERVER=...).
+                Note that server must be inserted protocol-less because services are using different protocols.
+                Another option is inserting URL addresses od these servers to file config.json
```
janinko: /od/of/
But I would write it so that the config.json is understood as the main way to configure and the ENV variable is alternative.

docs/users-documentation.html
```diff
@@ -164,16 +152,37 @@
...
                 </p>
             </li>
+            
+            <li>
+                After unpacking/clonning CLI tool, you need to configure servers.
+            <p>
+                You have to specify DA server and KeyCloak authentication server. DA server can be set as environmental variable ($ export DA_SERVER=...).
+                Note that server must be inserted protocol-less because services are using different protocols.
+                Another option is inserting URL addresses od these servers to file config.json 
+                DA server should be filled into "daServer" field. This server has to support WebSocket and HTTP protocol. Keycloak server should be in field "keycloakServer".
```
janinko: After updating to config.json structure, don't forget to update the doc here.

docs/users-documentation.html
```diff
@@ -164,16 +152,37 @@
...
+                In case one of these server is not specified, script exits with fail message.
+            </p>
+            </li>
+            
+            <li>
+                Before first run
+            <p>
+                Ensure that you have installed python3, websockets and asyncio packages.
```
janinko: I was also missing requests (or something like this) package.

docs/users-documentation.html
```diff
@@ -164,16 +152,37 @@
...
+            </p>
+            If you do not have them installed, it can be easily done by these commands:
+                <ul>
+                    <li>python instalation - <a href="https://www.unixmen.com/howto-install-python-3-x-in-ubuntu-debian-fedora-centos/">link</a></li>
+                    <li >$ pip3 install websockets</li>
+                    <li >$ pip3 install asyncio</li>
```
janinko: I think on Fedora it's better to use dnf to instal it to system.

docs/users-documentation.html
```diff
@@ -380,45 +389,65 @@
...
+        
+        
+        
+        <li>
+        <span class="strong">Administration tools</span>
+                <ul>
+                <li>
+                    These tools add, remove or change data. These are administrative tools, so you need to be authenticated to use them.
+                    You can authenticate with your credentials when you run admin tool. You are asked to enter your login. If no is entered, 
+                    your system user name is used. After you are asked to enter password. In case credentials are wrong, CLI tool exits with failure message with no data changed.
```
janinko: "If no is entered, your system user name is used."
I would rephrase this sentence. Maybe saying that the system user name is default?

docs/users-documentation.html
```diff
@@ -164,16 +152,39 @@
...
                         </ol>
                 </p>
             </li>
+            
+            <li>
+                After unpacking/clonning CLI tool, you need to configure servers.
+            <p>
+                You have to specify DA server and KeyCloak authentication server. It can be done by inserting URL addresses of these servers to file config.json 
+                DA server should be filled into "daServer" field. This server has to support WebSocket and HTTP protocol. Keycloak server should be in field "keycloakServer".
+                There are also "keycloakUser" and "keycloakRealm" fields, which should be filled by your credentials for proper working scripts.
```
janinko: keycloakUser field does not exist. Also it's not users credentials, it's ... well let's say keycloak configuration.

PR #397, opened 2016-12-05
[NCL-2700] Fixed error while bad response from WS

stanlyDoge:


cli-wrap/src/main/resources/cli/listings.py
```diff
@@ -528,7 +528,11 @@ def prod_difference():
             output = json.loads(output)["result"]
             return output
         except KeyError:
-            print(json.loads(output)["error"]["message"] + " - " + json.loads(output)["error"]["data"])
+            print(json.loads(output)["error"]["message"])
+            try :
+                print(json.loads(output)["error"]["data"])
+            except KeyError:
+                exit()
```
janinko: can you rather check for existence of the "data" field instead of exception throwing/catching?

stanlyDoge: yes

PR #399, opened 2016-12-05
Changed cofiguration file

stanlyDoge:


cli-wrap/src/main/resources/cli/config.json
```diff
@@ -1,4 +1,6 @@
 {
+    "keycloakUser" :"",
```
janinko: it's not user, it's client id

PR #401, opened 2016-12-06
Fixed executing DA-CLI from arbitrary folder

stanlyDoge:


janinko:
2 changes in one commit?

stanlyDoge:
renaming is related to import which has been modified so did not consider it as appropriate to split to two commits

cli-wrap/src/main/resources/cli/da-cli.py
```diff
@@ -1,7 +1,13 @@
...
 import sys
-from listings import *
 import argparse
+import inspect
+import os
+
+cmd_folder = os.path.realpath(os.path.abspath(os.path.split(inspect.getfile( inspect.currentframe() ))[0]))
+if cmd_folder not in sys.path:
+     sys.path.insert(0, cmd_folder)
+
```
janinko: Why are you inserting the path to sys.path? can't it be done just by prepending the path to config file name?

janinko:
and why did you renamed the import?

cli-wrap/src/main/resources/cli/da-cli.py
stanlyDoge: because other script files has some custom files import too and code to reach import them is ugly

stanlyDoge:
I renamed import because it was named pm (python module) which is not very good describing. Should I revert this change?

PR #406, opened 2016-12-07
installation script for needed packages

stanlyDoge:


cli-wrap/src/main/resources/cli/install.sh
```diff
@@ -0,0 +1,23 @@
...
+#tar -zxvf Python-3.5.2.tgz 
+#cd Python-3.5.2
+
+#./configure --prefix=/opt/python3 
+#make 
+#sudo make install
+
+#sudo dnf install python3-requests
+#sudo dnf install python3-websockets
+
```
jsenko: python3 may be already present, i would find if there exists python3.5 command, and then download and use ./configure without the prefix which will install python3.5
Also use pip instead of dnf
sudo pip3.5 install requests
sudo pip3.5 install websockets

janinko: Why would you use pip instead of dnf?

janinko:
Python 3 can be installed by dnf install python3

PR #422, opened 2017-01-12
[NCL-2774]: Set timeout to HTTP request to obtain maven-metadata.xml …

jbartece:
…from Indy

### DIFF
--- a/common/src/test/resources/testConfig.json
+++ b/common/src/test/resources/testConfig.json
@@ -21,7 +21,7 @@
           "aproxGroupPublic": "aprox-group-public",
           "backupScmUrl": "backup-scm-url",
           "cartographerServerUrl": "cartographer-server-url",
-          "defaultHttpRequestTimeout": 100000
+          "aproxRequestTimeout": 100000
         }
       ]
     }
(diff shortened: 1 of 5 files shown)


common/src/test/resources/testConfig.json
```diff
@@ -20,7 +20,8 @@
           "aproxGroup": "aprox-group",
           "aproxGroupPublic": "aprox-group-public",
           "backupScmUrl": "backup-scm-url",
-          "cartographerServerUrl": "cartographer-server-url"
+          "cartographerServerUrl": "cartographer-server-url",
+          "defaultHttpRequestTimeout": 100000
```
janinko: why is the property named with default?

jbartece: Because I was thinking about the situation, when we want to use set it also to other requests. If you want I can rename it

janinko: I would name it just httpRequestTimeout or aproxRequestTimeout.

jbartece: OK. I'll rename it to aproxRequestTimeout

janinko:
retest this please

PR #431, opened 2017-05-02
[NCL-2689] Improve DA requests performance 

janinko:
I introduced new ``ProductProvider`` interface, that unifies and abstract ways how we get artifact information.
This new interface is using Futures to enable more concurrent processing of results from different sources of product information.


reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
```diff
@@ -0,0 +1,230 @@
...
+    public static CompletableFuture<Set<ProductArtifacts>> filterProducts(
+            CompletableFuture<Set<ProductArtifacts>> artifacts,
+            Predicate<? super Product> predicate) {
+        return artifacts.thenApply(as -> as.stream()
+                .filter(a -> predicate.test(a.getProduct()))
+                .collect(Collectors.toSet()));
+    }
+
+    // TODO: filter unknown products, so that when there is artifact both in unknown product and in
+    // known product, it's removed from the unknown
```
jbartece: Is this todo left here intentionally? Won't it introduce regression?

bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
```diff
@@ -154,7 +168,7 @@ private void setDependencies(ProjectHiearchy hiearchy) {
     }
 
     private Set<ProjectHiearchy> toProjectHiearchies(Collection<GAV> gavs) {
-        return gavs.parallelStream().map(dep -> toProjectHiearchy(dep)).collect(Collectors.toSet());
+        return gavs.stream().map(dep -> toProjectHiearchy(dep)).collect(Collectors.toSet());
```
jbartece: Why did you change it to a standard stream?

reports-backend/src/main/java/org/jboss/da/products/api/Artifact.java
```diff
@@ -0,0 +1,15 @@
+package org.jboss.da.products.api;
+
+import org.jboss.da.model.rest.GAV;
+
+import lombok.Data;
+
+/**
+ *
+ * @author Honza Brázdil <janinko.g@gmail.com>
```
matedo1: Just reminder to change your email address.

reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
janinko: I was not entirely sure about it, and it was not critital to do. It should just enhance user experience so they wouldn't see:
```
foo:bar:1.0.0.redhat-1 built in EAP:1.0, EAP:1.1, UNKNOWN:UNKNOWN
foo:baz:3.2.1.redhat-1 built in UNKNOWN:UNKNOWN
```
but just
```
foo:bar:1.0.0.redhat-1 built in EAP:1.0, EAP:1.1
foo:baz:3.2.1.redhat-1 built in UNKNOWN:UNKNOWN
```

bc-backend/src/main/java/org/jboss/da/bc/impl/ProjectHiearchyCreator.java
janinko: toProjectHiearchy starts asynchronous tasks and adds them to the ``futures`` field. It then waits for them in the entry (public) methods.

reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
janinko: when we have an information that product was build in a specific product, we don't get any new information knowing it was build in unknown product (as it may be the specific one) and probably every time there is some specific, there is also unknown

PR #444, opened 2017-09-22
Duplicates

thescouser89:


thescouser89:
retest this please

thescouser89:
retest this please

reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
```diff
@@ -450,22 +449,19 @@ private BuiltReportModule toBuiltReportModule(GAV gav, VersionLookupResult vlr)
...
         request.getGavs().stream()
                 .distinct()
                 .forEach((gav) -> {
             LookupReport lr = new LookupReport(gav);
             reports.add(lr);
-            CompletableFuture<Set<ProductArtifacts>> artifacts = new CompletableFuture<>();
-            artifacts.supplyAsync(() -> gaProductArtifactsMap.get(gav.getGA())); //Need to check if it's correct
+
+            // get the CompletableFuture for the ga
+            CompletableFuture<Set<ProductArtifacts>> artifacts = gaProductArtifactsMap.get(gav.getGA());
```
jbartece: I think this approach, won't help us from reducing the number of calls to Indy. If I understand it correctly, for each GAV, we will do a request to Indy to get results of GA. I'm afraid, it's the same as before. To avoid the issues, we need to get the results for GA before, we do forEach on GAV.

jbartece:
Created a new PR based on the original code - https://github.com/project-ncl/dependency-analysis/pull/446

PR #446, opened 2017-09-26
Rework requests to Indy in lookup/gavs not to be duplicated for the same GA

jbartece:

PR #448, opened 2017-09-29
Rework requests to Indy in lookup/gavs not to be duplicated for the same GA - 1.2

jbartece:
Is already merged to 1.1 branch #446


janinko:
See comments in the other PR

PR #447, opened 2017-09-29
Rework requests to Indy in lookup/gavs not to be duplicated for the same GA - master

jbartece:
Is already merged to 1.1 branch https://github.com/project-ncl/dependency-analysis/pull/446

PR #451, opened 2017-10-19
Rework requests to Indy in lookup/gavs not to be duplicated for the same GA - master

thescouser89:
This PR is related to #447, applying changes based on Honza's comments in the PR.

* The last 4 commits in #447 has been squashed.
* The test was rewritten to explicitly list the unique GAVs we are expecting, instead on relying on `HashSet` to find unique GAVs (this might get broken if `hashCode` and `equals` implementation in the `GAV` class is broken)


thescouser89:
I had to do an additional commit for code clean + test improvement because it was hard to properly squash my changes into the right commit 😄

thescouser89:
Will use my execute powers and merge it. Will take responsibility for any bugs!

PR #456, opened 2017-11-29
Add the ability to add to a product whitelist from a file of GAV's.

jcoleman0redhat0com:


janinko:
In lookup command we have 
```
./da-cli.py lookup [GROUP_ID:ARTIFACT_ID[:PACKAGING[:CLASSIFIER]]:VERSION[:SCOPE]] 
```
so the GAV is optional and you can call it like
```
./da-cli.py lookup < gav-file
```

What do you think about using the same approach for black/product lists?

jcoleman0redhat0com:
That would also work for me.

jcoleman0redhat0com:
Changes updated to read the GAV list from stdin.

cli-wrap/src/main/resources/cli/da-cli.py
```diff
@@ -59,6 +59,8 @@ def printUsage(self):
         message += ("\n")
         message += ("    python " + sys.argv[0] + " add white GROUP_ID:ARTIFACT_ID:VERSION PRODUCT_NAME:VERSION\n")
         message += ("        Add artifact GROUP_ID:ARTIFACT_ID:VERSION to white list for a particular product\n")
+        message += ("    python " + sys.argv[0] + " add white PRODUCT_NAME:VERSION\n")
+        message += ("        Add artifact G:A:Vs from standard input to white list for a particular product\n")
```
janinko: I would just change the previous entry to read:
```
        message += ("    python " + sys.argv[0] + " add white [GROUP_ID:ARTIFACT_ID:VERSION] PRODUCT_NAME:VERSION\n")
        message += ("        Add artifact GROUP_ID:ARTIFACT_ID:VERSION to white list for a particular product.\n")
        message += ("        When G:A:V is not specified, reads list of G:A:Vs from standard input.\n")

```

PR #462, opened 2018-01-08
[NCL-3568] add option to override version suffix in lookup report

janinko:


reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
```diff
@@ -35,7 +35,7 @@
  *
  * @author Honza Brázdil &lt;jbrazdil@redhat.com&gt;
  */
-@ApplicationScoped
+//@ApplicationScoped // Should be Request scoped
```
thescouser89: Stupid question: does injection work when there are no annotation on the class?

common/src/main/java/org/jboss/da/common/version/VersionParser.java
```diff
@@ -9,10 +9,23 @@
 @ApplicationScoped
 public class VersionParser {
 
-    private static final String PATTERN_SUFFIX_BUILT_VERSION = "[.-]redhat-(\\d+)";
+    private static final String DEFUALT_SUFFIX = "redhat";
```
thescouser89: Typo on DEFAULT

reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
janinko: Yup. The default scope is [Dependent](https://docs.jboss.org/cdi/api/1.0/javax/enterprise/context/Dependent.html), which works here now, but the Request scope would be better. However Request scope doesn't work with websockets.

thescouser89: Alright! I haven't found any doc on that, but I guess it's not even documented(?)

thescouser89:
LGTM (minus tiny typo)! 👍

reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
janinko: https://docs.oracle.com/javaee/6/tutorial/doc/gjbbk.html
> Dependent
> @Dependent
> The default scope if none is specified; it means that an object exists to serve exactly one client (bean) and has the same lifecycle as that client (bean).

PR #464, opened 2018-01-26
[NCL-3604] Add retry for status 500 from Indy

thescouser89:
This commit also adds a retry for HTTP status 500 from Indy when
fetching the maven-metadata.xml file. This is in addition to the current
logic which is to retry on HTTP status 504.

This is done as Indy sometimes recovers from HTTP status 500 errors to
provide us with the right content (See NOS-1155).

This commit also adds a waiting time between retries; with the waiting
time determined by using an exponential backoff. This will hopefully
help the network/Indy recover from overload and provide the correct data
at a later time.


thescouser89:
@jbartece Sure I can change the retries from 5 to 2

PR #466, opened 2018-02-06
[NCL-3465] Add read timeout when communicating with Indy

janinko:
We had only ConnectTimeout set but that's only for initial socket
connection, we also need read timeout if we want to timeout when indy
takes a long time to respond.


janinko:
retest this please

PR #468, opened 2018-02-28
[NCL-3623] Get list of available versions for NPM artifacts from Indy

janinko:


thescouser89:
Oh only 927 files changed. Perfect timing for vacation :smile: 

Have fun @jbartece !

janinko:
Check just the first 3 commits, the 4th is just the same change in all the 900 files

communication/src/test/java/org/jboss/da/communication/aprox/AproxConnectorTest.java
```diff
@@ -40,13 +40,16 @@
     @Mock
     private Logger log;
 
+    @Spy
+    private MetadataFileParser parser = new MetadataFileParser();
```
jbartece: Why adding @Spy and also initializing the entity?

janinko: @Spy can be used only on initialized fields. The @Spy was put here so the @InjectMocks would recognize it and inject it, but since it's not injected anymore I forgot to remove it from here too.

SPY: I'm OK with uninitialized fields as well.

janinko:
Merged into feat-npm branch. Will merge the feat-npm branch to master when all the NPM support is ready.

PR #476, opened 2018-06-13
[feature] Change Artifact to interface and extract MavenArtifact

janinko:


reports-backend/src/main/java/org/jboss/da/reports/impl/ReportsGeneratorImpl.java
```diff
@@ -206,7 +206,7 @@ private void traverseAndFill(ArtifactReport report, Set<Product> products,
         return availableArtifacts.thenApply(pas-> {
             List<String> versions = pas.stream()
                 .flatMap(as -> as.getArtifacts().stream())
-                .map(a -> a.getGav().getVersion())
+                    .map(a -> a.getVersion())
                     .collect(Collectors.toList());
```
jbartece: Fix code style

PR #482, opened 2018-08-23
[NCL-3622] Create new NPM alignment endpoint

janinko:


janinko:
@jbartece OK to review

PR #483, opened 2018-08-30
[NCL-4079] When version suffix is provided, use both provided and default

janinko:
When different version suffix is used, it's usead as additional (priority)
suffix, not as an override for the default one.

suffix: null
 returned: redhat-1, redhat-2, ...
 bestmatch: redhat-2

suffix: foo
 returned: foo-1, redhat-1, redhat-2, ...
 bestmatch: foo-1

### DIFF
--- a/common/src/test/java/org/jboss/da/common/version/VersionAnalyzerTest.java
+++ b/common/src/test/java/org/jboss/da/common/version/VersionAnalyzerTest.java
@@ -155,12 +155,51 @@ public class VersionAnalyzerTest {
     }
 
     private void checkBMV(String expectedVersion, String version, String[] versions) {
-        VersionAnalyzer.VersionAnalysisResult result = versionFinder.analyseVersions(version,
+        checkBMV(versionFinder, expectedVersion, version, versions);
+    }
+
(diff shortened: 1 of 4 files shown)


common/src/test/java/org/jboss/da/common/version/VersionAnalyzerTest.java
```diff
@@ -163,4 +163,68 @@ private void checkBMV(String expectedVersion, String version, String[] versions)
...
+        String[] avaliableVersions;
+        VersionAnalyzer.VersionAnalysisResult result;
+        Optional<String> bmv;
+
+        avaliableVersions = new String[] { "1.4.0.redhat-1", "1.4.0.temporary-redhat-1" };
+        result = versionAnalyzer.analyseVersions(version, Arrays.asList(avaliableVersions));
+        bmv = result.getBestMatchVersion();
+        assertEquals(2, result.getAvailableVersions().size());
+        assertTrue("Best match version expected to be present", bmv.isPresent());
+        assertEquals(expectedVersion, bmv.get());
```
jbartece: Reduce the code repetition please

PR #491, opened 2019-07-16
NCL-4972 - Add metric to track errors in rest endpoints

vibe13:
### Checklist:

* [ ] Have you added a note in the CHANGELOG.md for your change if user-facing?


thescouser89:
@vibe13 off topic, but one thing I've been struggling is to port the pncmetrics dependency to Causeway / DA to stop the duplication. If you are good with the ear configuration, could you maybe try to use the dependency instead of the module? I tried but failed miserably :(

PR #521, opened 2020-09-15
[NCL-6030] Remove testsuite activation by profile

jomrazek:
### Checklist:

* [ ] Have you added a note in the CHANGELOG.md for your change if user-facing?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -442,29 +442,6 @@
     </build>
     <profiles>
         <profile>
-            <id>excludeTests</id>
-            <activation>
-                <property>
-                    <name>testsuite.test.excludes</name>
-                </property>
-            </activation>
-            <modules>
-                <module>testsuite</module>
-            </modules>
-            <build>
-                <plugins>
-                    <plugin>
-                        <artifactId>maven-failsafe-plugin</artifactId>
-                        <configuration>
-                            <excludes>
-                                <exclude>${testsuite.test.excludes}</exclude>
-                            </excludes>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-        <profile>
             <id>excludeRemoteTests</id>
             <activation>
                 <property>
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,7 @@
         <module>reports-rest</module>
         <module>application</module>
         <module>cli-wrap</module>
+        <module>testsuite</module>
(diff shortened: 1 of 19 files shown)


jomrazek:
:exclamation::exclamation::exclamation:When merged change jenkins build script to
`mvn clean install -DtestsuiteContainer=$PATH_TO_EAP -Dtestsuite.test.excludes=**/BugReporoducerRemoteTestIT.java`

pom.xml
```diff
@@ -424,22 +425,45 @@
...
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
 
     </build>
     <profiles>
         <profile>
-            <id>testsuite</id>
+            <id>excludeTests</id>
```
matejonnet, line 445: What's the use case of this profile ?

jomrazek: To exclude any test, it's used in jenkins for example (not sure if still needed) I left it as it was just moved it so remote test exclusion is actually working

matejonnet: Don't know about this use case but in general if the test is failing it's better to ignore it on the test source level.
If it's special type of test that needs to be excluded by default an option is to create a "group" and exclude it.

It's good to have a project in s state that it builds and all the tests pass for anyone with standard `mvn clean install` command.

jomrazek: I have no problem to ignore class that's excluded in jenkins and remove this profile, but as I said I don't know reasoning behind this approach

jomrazek:
:exclamation:  remove 
`-Dtestsuite.test.excludes=**/BugReporoducerRemoteTest.java` from jenkins PR job after merge

@janinko you might comment on tests that were excluded in past if you have some concerns about ignoring them now

janinko:
>  remove
> `-Dtestsuite.test.excludes=**/BugReporoducerRemoteTest.java` from jenkins PR job after merge
> 
> @janinko you might comment on tests that were excluded in past if you have some concerns about ignoring them now

The test was ingored only in PRs as it took long time to run. I'm not sure if that's still the case when we use wiremock, maybe we should try to un-ignore the test.

jomrazek:
@janinko those excluded tests are failing now

jomrazek:
Ignored tests, they were not performed from beginning and I think is out of scope of this issue to fix them when they're failing

janinko (approved):

PR #523, opened 2020-11-05
[NCLSUP-132] Allow dash instead of dot as separator before MICRO in version parser

janinko:
### Checklist:

* [x] Have you added a note in the [CHANGELOG](https://github.com/project-ncl/dependency-analysis/wiki/Changelog) for your change if user-facing?
* [x] Have you added unit tests for your change?


rnc (approved):
While I'm still not very happy at you having to change DA and this discrepancy appearing - which seems to have been historical from years back, unless we unify or create a common version library this might continue to happen. I'll put an issue into PME for David to review its handling in future to make sure we are the same around this.

PR #527, opened 2021-02-11
Managedsvc

janinko:
### Checklist:


janinko:
@pkocandr @thescouser89 Sorry guys, didn't know who else to pick for this beast.
It may be easier to review commit by commit then everything at once.

janinko:
retest this please

janinko:
retest this please

thescouser89:
It's such a big PR :( I don't think I have the mental capacity to be able to review it. I'll try. Are there any specific commits you'd like me to look at?

janinko:
- Remove lombok experimental onMethod as it behaves wird on different JDK
- Fix wrong return type in apidocs
  - These two ^^ are actually not really related, just 2 fixes.
- Chnage version classes to work with list of suffixes
  - In this ^ commit, the main logic was changed. This is the most critical commit.
- Update reports-backend to new version parsing
  - This ^ updates the project to corectly use what was done in previous commit (and compile)
- Use PNC Product Provider for all requests
- Update wiremock stubs with PNC communication and update expected repsonses apropriately
  - Most of this ^ commit are (re)generated stub files and updated data in expected test results. Non-java files can be ignored
- Use only one repository when doing indy lookup
- Introduce JAX-RS Exception mapper
  - This ^ is improving exception handling in REST. Mostly by moving three methods to different place.
- Introduce new configuration Lookup Mode
  - This ^ is one of the more important commits
- Update wiremock stubs
  - This ^ can be ignored
- rename org.jboss.da.rest.spi to api
  - This ^ can be ignored
- Create API for new lookup endpoints 
  - Just adding new API endpoint methods and DTOs
- Extract Futures util class
  - Just a move of 2 methods
- Return CompletableFuture from ProductProvider methods
  - Nothing exiting
- Add implementation of Lookup Generator
  - This ^ is one of the more important commits, introducing business logic for the new endpoints
- Add new lookup endpoints implementation
  - Nothing exiting
- Tests for new lookup endpoints
- Add Jboss snapshot repository to pom to pom

reports-model/src/main/java/org/jboss/da/lookup/model/MavenLookupRequest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @NonNull
+    private final Set<GAV> artifacts;
+
+    @NonNull
+    private final String mode;
+
+    @NonNull
+    @JsonProperty(defaultValue = "false")
+    private final boolean brewPullActive;
```
janinko, line 32: boolean can't be null, annotation not necessary

thescouser89: are you reviewing your own pr?

janinko: you don't? :nerd_face:

reports-model/src/main/java/org/jboss/da/lookup/model/NPMLookupRequest.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @NonNull
+    private final Set<NPMPackage> artifacts;
+
+    @NonNull
+    private final String mode;
+
+    @NonNull
+    @JsonProperty(defaultValue = "false")
+    private final boolean brewPullActive;
```
pkocandr, line 27: We don't need brewPullActive for NPM (for now), because it is not implemented on Indy side. Brew pull implementation allows to query only Maven artifacts and AFAIK there is no request to implement that for NPM packages.

pkocandr (changes requested):
Very nice work, although as descussed earlier today it should have been split into multiple PRs. Just a few changes requested down there. Feel free to refuse them if you think they are not needed.

common/src/main/resources/da-config.json
```diff
@@ -17,7 +17,21 @@
...
+              "pncCategory": "STANDARD",
+              "pncQualities": [ "NEW", "VERIFIED","TESTED"]
+            },
+            {
+              "name": "TEMPORARY",
+              "suffixes": ["temporary-redhat", "redhat" ],
+              "pncCategory": "STANDARD",
+              "pncQualities": [ "NEW", "VERIFIED", "TESTED", "TEMPORARY" ]
+            }
+          ]
```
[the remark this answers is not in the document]
janinko: I don't think so. This file is not used at all in production, it's just a template.

janinko:
retest this please

common/src/main/java/org/jboss/da/common/version/VersionAnalyzer.java
```diff
@@ -17,62 +17,58 @@
...
-                .distinct()
-                .peek(v -> sortedVersions.add(v))
-                .map(versionParser::parse)
-                .collect(Collectors.toList());
-
-        SuffixedVersion version = versionParser.parse(querry);
-        Optional<String> bmv = findBiggestMatchingVersion(version, parsedVersions);
-
-        return new VersionAnalysisResult(bmv, sortedVersions);
+        List<String> sortedVersions = versions.stream().sorted(comparator).distinct().collect(Collectors.toList());
```
vibe13, line 47: Could be changed to versions.stream().distinct().sorted(comparator).collect(Collectors.toList()); maybe? (distinct before sorting)

common/src/main/java/org/jboss/da/common/version/VersionAnalyzer.java
```diff
@@ -86,38 +82,40 @@ public VersionAnalysisResult analyseVersions(String querry, Collection<String> v
...
-        if (firstMatcher.group("minor") == null && secondMatcher.group("minor") != null) {
-            return second;
+        boolean firstIsOSGi = first.equals(VersionParser.getOSGiVersion(first));
+        String firstMinor = firstMatcher.group("minor");
+        String firstMicro = firstMatcher.group("micro");
+        boolean returnFirst;
+
+        if (firstIsOSGi != second.equals(VersionParser.getOSGiVersion(second))) {
+            returnFirst = firstIsOSGi; // One of the version is not OSGi, prefer the OSGi version
+        } else if (!Objects.equals(firstMinor, secondMatcher.group("minor"))) {
```
vibe13, line 105: Not sure I understand this, couldn't it be that e.g. firstMinor is 3 and secondMinor is 5? The comment seems to contemplate only the presence of a null value as the reason why the !Objects.equals is true

common/src/test/java/org/jboss/da/common/version/VersionComparatorTest.java
```diff
@@ -121,6 +132,19 @@ public void testBaseVersion() {
         assertTrue(vc.compare("2.2.0.MR1", "2.2.0.CR1") < 0);
     }
 
+    @Test
+    public void testBaseVersionWithMultipleSuffixes() {
+        VersionComparator vc1 = new VersionComparator("3.4.2.Final", new VersionParser("temporary-redhat", "redhat"));
+
+        assertTrue(vc1.compare("3.4.2.Final", "3.4.2.Final-redhat-1") < 0);
```
vibe13, line 139: Sorry, not sure I understand, which version of the 2 "wins"? is it the first one or the second?  If it's the first I understand e.g. vc1.compare("3.4.2.Final-redhat-2", "3.4.2.Final-redhat-1") < 0 but not vc1.compare("3.4.2.Final", "3.4.2.Final-redhat-1") < 0

reports-backend/src/main/java/org/jboss/da/reports/impl/LookupGeneratorImpl.java
```diff
@@ -0,0 +1,163 @@
...
+
+    private Map<String, CompletableFuture<Set<String>>> getProductArtifacts(Set<NPMPackage> packages) {
+        return packages.stream()
+                .map(NPMPackage::getName)
+                .distinct()
+                .map(name -> new NPMArtifact(name, "0.0.0"))
+                .collect(Collectors.toMap(Artifact::getName, pncProductProvider::getAllVersions));
+    }
+
+    private CompletableFuture<Set<ProductArtifacts>> filterBlacklistedArtifacts(
```
vibe13, line 109: Where is it used?

vibe13:
@janinko I have left a few comments, otherwise impressive job, thanks!

common/src/main/java/org/jboss/da/common/version/VersionAnalyzer.java
janinko: We actually don't expect that there will be duplicate versions, so no expected gain in sorting time if we first remove duplicates.
On the other hand having distinct after sorted makes the distinct calculation faster, as it doesn't need to construst and populate hash set.
(I didn't know this before your comment :D)

common/src/main/java/org/jboss/da/common/version/VersionAnalyzer.java
janinko: From the method documentation:
```java
    /**
     * Assuming the two versions have the same OSGi representation, returns the more specific version. That means
     * X.Y.Z.something is preffered to X.Y.something which is preffered to X.something.
     */
```
So it compares only versions that are "logically the same", so there can't be different number in firstMinor and secondMinor.
(This also means that if firstMinor is 3, then secondMinor is also always 3 and never null, because null can be "replaced" only by 0.)

common/src/test/java/org/jboss/da/common/version/VersionComparatorTest.java
janinko: You can read:
```
assertTrue(vc1.compare("3.4.2.Final", "3.4.2.Final-redhat-1") < 0);
```
as if you move the inequality sign between the strings:
```
"3.4.2.Final" < "3.4.2.Final-redhat-1"
```
i.e. "3.4.2.Final" is closer to the "base" version "3.4.2.Final" (which is true because they are equal :D)

```
"3.4.2.Final-redhat-2" < "3.4.2.Final-redhat-1"
```
"3.4.2.Final-redhat-2" wins here because if the versions are the same except for the suffix number, highest suffix number wins.

These comparsions to the base are for suggesting the best replacement for users. And in practice only compare versions with suffix.

reports-backend/src/main/java/org/jboss/da/reports/impl/LookupGeneratorImpl.java
janinko: Good question

common/src/main/java/org/jboss/da/common/version/VersionAnalyzer.java
vibe13: ah ok. good then :-)

common/src/main/java/org/jboss/da/common/version/VersionAnalyzer.java
vibe13: aaah got it now. thanks!

common/src/test/java/org/jboss/da/common/version/VersionComparatorTest.java
vibe13: ook, I missed the "is closer to the "base" version" part. Understood now, thanks!

pkocandr:
Why are you force pushing changes? It is impossible to see what are the changes since my last review, so I have no idea if my concerns were addressed or not :-(

pkocandr (approved):

vibe13:
+1 for me!

PR #530, opened 2021-03-26
Prepare release 2.1.0 without releasing PNC

pkocandr:
### Checklist:


pkocandr:
Giving up, we are going to do a partial release of just parent and reports-model, which does not depend on PNC.

PR #532, opened 2021-05-11
Changes to better deserialize JSON using the DTO

thescouser89:
This change applies to:

- LookupReport
- NPMLookupReport

The changes are required to properly deserialize the JSON into the
objects for clients using the DTOs to deserialize.

While a few stuff were tried, it looks like Jackson is confused by the
combination of the 'Builder' pattern and '@JsonUnwrapped' annotation and
doesn't know what to do. To keep it simple, this commit removes the
`@Builder` annotation, and uses the 'NoArgsConstructor' to make sure
Jackson can deserialize properly.

Tests are written to make sure that the deserialization process works.

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -311,6 +311,11 @@
                 <artifactId>wiremock-standalone</artifactId>
                 <version>2.6.0</version>
             </dependency>
+            <dependency>
+                <groupId>org.assertj</groupId>
+                <artifactId>assertj-core</artifactId>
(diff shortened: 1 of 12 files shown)


thescouser89:
Deploying it on our devel, I can also confirm Swagger is happy with the new annotations.

thescouser89:
@janinko Could you please have a look?

PR #536, opened 2021-06-14
New API for geting latest version

janinko:
### Checklist:


rnc:
Looks good to me , I guess its possible to share common definitions between the lookup/latest request objects but they are such small classes its probably not really worth it.

PR #548, opened 2022-10-17
[NCL-7184] Intoduce endpoint for avalable versions lookup

janinko:
### Checklist:

### DIFF
--- a/reports-backend/src/main/java/org/jboss/da/reports/impl/LookupGeneratorImpl.java
+++ b/reports-backend/src/main/java/org/jboss/da/reports/impl/LookupGeneratorImpl.java
@@ -8,7 +8,10 @@ import org.jboss.da.common.version.VersionAnalyzer;
 import org.jboss.da.listings.api.service.BlackArtifactService;
 import org.jboss.da.lookup.model.MavenLatestResult;
 import org.jboss.da.lookup.model.MavenLookupResult;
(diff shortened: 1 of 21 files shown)


reports-backend/src/main/java/org/jboss/da/reports/impl/LookupGeneratorImpl.java
```diff
@@ -69,17 +72,32 @@ public LookupGeneratorImpl(Configuration config) throws ConfigurationParseExcept
...
 
+    @Override
+    public Set<MavenVersionsResult> lookupVersionsMaven(
+            Set<GAV> gavs,
+            VersionFilter vf,
+            String mode,
+            boolean brewPullActive,
+            boolean includeBad) throws CommunicationException {
+        LookupMode lookupMode = getMode(mode, includeBad);
+        System.out.println("Lookup mode: " + lookupMode + " include bad?" + includeBad);
```
jomrazek, line 89: I guess this is for debugging as it is just print and it's logged one level higher in LookupImpl?

jomrazek (approved):
one comment, other than that looks good to me

thescouser89 (approved):
I think it looks ok to me. Spent some time parsing what is `v`, `vf` and `vc` but code for comparison looks good and their are unit tests and integration tests backing it up.

PR #570, opened 2023-06-20
Remove Whitelist

janinko:
### Checklist:


reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
```diff
@@ -66,10 +65,6 @@ public static CompletableFuture<Set<ProductArtifacts>> filterProducts(
     // TODO: filter unknown products, so that when there is artifact both in unknown product and in
```
michalovjan, line 65: is this todo relevant still?

michalovjan (changes requested):

reports-backend/src/main/java/org/jboss/da/products/impl/AggregatedProductProvider.java
janinko: yup, potentionally

michalovjan: Decision for the future.

michalovjan (approved):

PR #591, opened 2025-08-05
Create super-classes for lookup results

jomrazek:
This change should help to eliminate order methods for every result type in bacon https://github.com/project-ncl/bacon/blob/main/da/src/main/java/org/jboss/bacon/da/DaHelper.java#L203


michalovjan (approved):

akridl (approved):

akridl:
not sure whether you needed my review still even after Ján approved it, but here you go!! :smile:

thanks for simplification of the code!

PR #649, opened 2026-04-01
NCL-9588 - Move configuration from json to quarkus yaml

rnc:
This is a fairly large change migrating the configuration to Quarkus native (with the assistance of Cursor!). This will break current deployments so @thescouser89 will have to fix up NCL-9575


thescouser89 (approved):
I didn't reviewed everything, but looks like the codebase is cleaner now with removing the configuration part from PNC. I like it!

jomrazek (approved):
looks good

PR #661, opened 2026-04-10
NCL-9576 Metrics

rnc:
### Checklist:

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -54,6 +54,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-elytron-security-ldap</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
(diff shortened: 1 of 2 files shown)


application/pom.xml
```diff
@@ -54,6 +54,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-elytron-security-ldap</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-opentelemetry</artifactId>
+    </dependency>
```
rnc, line 60: @thescouser89  Would it be better to use https://quarkus.io/guides/telemetry-micrometer-to-opentelemetry ?

thescouser89: still tech preview though :(

I think having the micrometer and opentelemetry both installed is fine for now.

thescouser89 (approved):

PR #710, opened 2026-06-18
[NCLSUP-1440] Fix transaction rollback error in lookup endpoints

thescouser89:
Move @Transactional from REST layer to service layer to prevent transaction suspension issues.

Changes:
- Remove @Transactional from LookupImpl REST endpoint class
- Add @Transactional to BlackArtifactServiceImpl methods that perform database operations (prefetchGAs, addArtifact, getArtifact, removeArtifact, getArtifacts)

The transaction timeout is likely only 5 minutes. If the transaction starts at the REST endpoint level and Indy takes a long time to align, the transaction will already timeout by the end of the endpoint and fail.

This ensures database operations run in their own short-lived transactions while ProductProvider async operations execute completely outside of any transaction context, as designed.

### Checklist:


thescouser89:
The transaction timeout is actually just 1 minute! https://quarkus.io/guides/transaction#quarkus-narayana-jta_quarkus-transaction-manager-default-transaction-timeout

rnc (approved):

PR #712, opened 2026-06-19
Fix file handle leak in LocalRepo by closing Files.walk() streams

thescouser89:
Files.walk() returns a Stream<Path> backed by DirectoryStream which holds file handles that must be closed. Failure to close these streams can lead to file descriptor exhaustion ("Too many open files" errors). Wrapped both getAllPoms() methods with try-with-resources to ensure proper cleanup.

### Checklist:


jomrazek (approved):
Nice! 👍

PR #727, opened 2026-06-30
[NCL-9724] Fix CVEs

patrikk0123:
### Checklist:


patrikk0123:
First migrated to maven scm provider v2, which brought more CVEs.
v2 also doesn't support cloning by hash, so made the implementation fully cover the git cloning, without the scm provider.
Now scm provider isn't needed for the git part.

rnc:
I've been looking at the original code and I hadn't realised it was also executing the git binary and only using maven-scm (and jgit) as a fallback. I'm not completely convinced by this in the sense the maven-scm stuff seems pointless - we don't support SVN so that is irrelevant now. I am not sure of the need to use maven-scm as opposed to using jgit directly. As for executing git, @akridl has a set of nice git execution functions in reqour that could be reused (and also he is using smallrye processexecutor which avoids pitfalls in the java process code. I would be tempted to add shallow clone to reqour, strip out all of the maven-scm code, and reuse reqour code here. ( cc @akridl  @thescouser89 )

patrikk0123:
@rnc 
you mean these?
https://github.com/project-ncl/reqour/blob/main/core/src/main/java/org/jboss/pnc/reqour/common/GitCommands.java
https://github.com/project-ncl/reqour/blob/main/core/src/main/java/org/jboss/pnc/reqour/common/utils/GitUtils.java

do you think it should be added to pnc-common? though, there is a lot of it

rnc:
@patrikk0123 Yes ; good question ; that probably should be discussed with @akridl and others offline on the best approach here. While your PR probably will fix the issue, I think it would be simpler to strip out all the maven-scm stuff entirely.

rnc (approved):

PR #733, opened 2026-07-09
Add 404 handling

rnc:
### Checklist:

### DIFF
--- a/communication/src/main/java/org/jboss/da/communication/repository/MetadataFileParser.java
+++ b/communication/src/main/java/org/jboss/da/communication/repository/MetadataFileParser.java
@@ -23,7 +23,7 @@ public class MetadataFileParser {
     @Inject
     ObjectMapper om;
 
-    public static VersionResponse parseMavenMetadata(InputStream in) throws JAXBException {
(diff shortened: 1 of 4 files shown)


communication/src/main/java/org/jboss/da/communication/repository/MetadataFileParser.java
```diff
@@ -23,7 +23,7 @@ public class MetadataFileParser {
     @Inject
     ObjectMapper om;
 
-    public static VersionResponse parseMavenMetadata(InputStream in) throws JAXBException {
+    public VersionResponse parseMavenMetadata(InputStream in) throws JAXBException {
```
michalovjan, line 26: Why does this need to be non-static?

michalovjan (changes requested):

communication/src/main/java/org/jboss/da/communication/repository/MetadataFileParser.java
rnc: It doesn't really ... everything was calling it as `parser.parseMavenMetadata` so either change the callers or change the function to silence the editor warnings.

